PR #1, opened 2019-12-27
[chg] avoid dealing with encoding and provide backwards compatibility

raw-data:
While running CAPEv2 with virtualbox machinery, the system enters in a loop state. It seems no match is found for the detected VM state when checked against the predefined set (e.g. POWEROFF, ABORTED, etc.); `setting universal_newlines=True` returns output as text instead of binary and it does the trick.

### DIFF
--- a/modules/machinery/virtualbox.py
+++ b/modules/machinery/virtualbox.py
@@ -72,6 +72,7 @@ class VirtualBox(Machinery):
 
         try:
             if subprocess.call(virtualbox_args,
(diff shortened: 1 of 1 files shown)


doomedraven:
amazing, thanks a lot, as we use only kvm, is the most tested hypervisor, so we appreciate a lot your collaboration, and sorry for delay in merge

raw-data:
No stress! Keep going with the great job!

PR #9, opened 2020-01-08
Fix Screenshots to work with Pillow versions newer than 6.1

wmetcalf:
https://github.com/python-pillow/Pillow/issues/4130


doomedraven:
thanks a lot

PR #10, opened 2020-01-08
Make image dedupe hash configurable. In my testing avg hash 'ahash' a…

wmetcalf:
…ctually provides the best balance between over/under prune. Also in testing doing a reverse sort of image list provides slightly more complete images. We resort after prune to put them back in the proper order.


doomedraven:
thanks!

PR #14, opened 2020-01-10
Add various signatures

ditekshen:
I haven't tested these against CAPEv2 since I don't have the capacity to run it yet. Please ignore the PR if this is going to be an issue, and will PR against CAPEv1. Sorry for the commits mess, I inadvertently overwritten an existing signature :)


doomedraven:
the signature should be more then transparent between version, the uniq change is python2 vs 3 syntax, so all fine, thanks :)

PR #19, opened 2020-01-16
Ignore innocuous keyerror when disposing db connection

enzok:
This will get rid of the ugly error trace


doomedraven:
heh nice trick thanks

PR #23, opened 2020-01-20
Proper memdump freespace fix for non-ramfs mode.

wmetcalf:


doomedraven:
thanks again, as im using tmpfs i totally forgot

PR #26, opened 2020-01-23
Add signature to detect malware encryption

ditekshen:
This covers two families for now. Hopefully more gets added.


doomedraven:
nice stuff

PR #30, opened 2020-01-27
Fix - Show filepath of dropped file

enzok:
instead of the hash


doomedraven:
i think if you use full path youneed to use |safe

enzok:
added |safe

PR #40, opened 2020-01-29
Fix CAPE.py bugs - read all file data, key check

enzok:
Fix PeID static processing
Code cleanup


doomedraven:
Hey thanks can you change last one to list(set( please?

PR #43, opened 2020-02-05
Fix screenshots download

aaguilartablada:
When you try to download all screenshots from an analysis in ZIP format an exception occurs:
string argument expected, got 'bytes'
This change solve the problem. Tested OK


doomedraven:
thanks

PR #44, opened 2020-02-06
Logging fixed to avoid TypeError

elmaisbuscado:
Those parameters cause this exception due to capital letter on backupCount variable.
--- Logging error ---
  File "/usr/lib/python3.6/logging/handlers.py", line 72, in emit
    self.doRollover()
  File "/usr/lib/python3.6/logging/handlers.py", line 397, in doRollover
    if self.backupCount > 0:
TypeError: '>' not supported between instances of 'NoneType' and 'int'


doomedraven:
thanks a lot i totally forgot about this one

PR #47, opened 2020-02-07
Add signature for disabling Folder Options menu

ditekshen:
Examples:
- https://www.microsoft.com/en-us/wdsi/threats/malware-encyclopedia-description?name=Win32%2FBrontok
- https://blog.malwarebytes.com/detections/hijack-folderoptions/


doomedraven:
thanks for another interesting signature

PR #62, opened 2020-02-14
Update analyzer.py

enzok:
Fix syntax error


doomedraven:
Lol i had tested that and didt spot, thanks

PR #71, opened 2020-02-21
Add ObliqueRAT signatures

ditekshen:
Reference: https://blog.talosintelligence.com/2020/02/obliquerat-hits-victims-via-maldocs.html


doomedraven:
thank you

PR #74, opened 2020-02-22
Add initial linux support - basic support, no behavior analysis

enzok:


kevoreilly:
Any chance we could make this something we can switch off in the config? The changes to the web ui are currently pretty grim looking for a windows only user...

doomedraven:
I will change that

PR #83, opened 2020-03-03
Delete this signature

naxonez:
Makes false positive: 
Example (Trickbot): f69ae399eacdd1f1112ee880a1c851bf3895400fd5e89e7ce1d215b2f1674ec3


ditekshen:
@naxonez can you point me where this the FP is generated with this sample? I can't see it anywhere I looked.

naxonez:
Hi,

I had the issue with the following trickbot sample:

https://capesandbox.com/analysis/13531/

With this signature, cape detects it as revil.


Regards

doomedraven:
ok i have recheked on my side that, it not giving me FP, naxonez is rechecking again on his side

ditekshen:
That's odd. The sample has curly braces surrounding the guid, which should not match. I check on my side as well and I did not see it FP. Maybe it is a different sample? I will keep an eye on this on my dev boxes. I would keep it deleted until proven otherwise. Thanks @naxonez for reporting this.

doomedraven:
i have reintroduced it

PR #101, opened 2020-03-10
Update rat_dcrat.py

ditekshen:


doomedraven:
thank you

doomedraven:
btw not sure if you noted but when you pushing different PRs, they are all in the same latest pull request

ditekshen:
I saw that but wasn't sure why. I'll do more reading on this

doomedraven:
not sure, but maybe bcz when you creating PR, you pushing to master and then pull request? you can change name from master to something else, but that just headups, as that doesn't hurt nothing at all :)

PR #106, opened 2020-03-12
Update plugins.py

firmianay:
'family' seems like a redundant variable

### DIFF
--- a/lib/cuckoo/core/plugins.py
+++ b/lib/cuckoo/core/plugins.py
@@ -289,14 +289,15 @@ class RunProcessing(object):
         else:
             log.info("Logs folder doesn't exist, maybe something with with analyzer folder, any change?")
 
-        family = ""
-        self.results["malfamily_tag"] = ""
         if self.results.get("detections", False):
-            family = self.results["detections"]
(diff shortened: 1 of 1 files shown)


doomedraven:
is what i had tried yesterday, but that isnot optimal solution https://github.com/kevoreilly/CAPEv2/commit/b66810ef22e24c5f39f42448ce40ec27a7fc6458, the idea is to not check the rest once it detected, and it going from more trusted "software" to lowest, as if we remove if not family it will enter each section and we don't want that, but thank you for your time :)

firmianay:
But now it also enter each section if `self.results.get("detections", False)` returns False. So I think `if not family` is useless.

doomedraven:
it should enter each section, if cape detection is empty, as it works as:
```
If CAPE
elif suricata
elif VT
elif ClamAV
```

that is our detection way to try to get malfamily by that software/services

doomedraven:
the thing here, is we check family,bcz it can not set, imagine that what that service detects, is blacklisted and family keeps empty, is why we walk that way, but if you have improvement idea is more than welcome

firmianay:
ok if you want to keep this control flow structure.

PR #146, opened 2020-03-27
Update index.html

firmianay:


doomedraven:
nice catch, as i did sed /malfamily/detections/ and didn't spot it

PR #157, opened 2020-04-01
Update views.py

firmianay:
`if not xxx` will always be Flase


doomedraven:
hello but i don't see any change related to that, only removed try/except which is required as inside of the search there is `raise`

PR #167, opened 2020-04-08
Adds chm package

bartblaze:
chm.py already exists - this simply adds the logic for detecting these files.


doomedraven:
thanks, in general it was opened in IE no?

bartblaze:
Nope, triggered the generic package for me (so cmd). CHM needs hh.exe to run properly :+1:

PR #173, opened 2020-04-15
Add calendar file

bartblaze:
This should work when it's opened by Outlook. In regards to:
http://www.pwncode.io/2020/04/outlook-calendar-file-ics-file-format.html


doomedraven:
but as far as i understand, user still should click on link no?

bartblaze:
Yes, correct. Not sure we can add that behaviour to the package. At least with this it will be opened in Outlook. :smiley: But from strings it's easy to check the link too tbh.

PR #179, opened 2020-04-18
Create Bokbot.yar

r0ny123:


doomedraven:
thanks

doomedraven:
FYI all yara rules that is written by community is moved to community repo

r0ny123:
Thnaks!
what do you mean by core yara and community yara?

doomedraven:
core yara, is done by @kevoreilly  or me, https://github.com/kevoreilly/community there is community stuff, so is that way we avoid to be responsable of FP, bad code, not maintained plugins that we don't use, so everyone can just remove them or do whatever they want on their side

doomedraven:
as example you posted yara, that won't be shown in webgui due to missed correct type inside of cape_type, do search by `cape_type` in repo and you will see the difference

r0ny123:
Thanks for the explanation!

doomedraven:
You are welcome, thanks for yara

PR #186, opened 2020-04-23
chg: Bump MISP reporting module

### LINKED ISSUE #184: Relying on a very outdated version of PyMISP
Someone mentioned an issue related to PyMISP using CAPEv2 and it seems that you're relying on an extremely outdated version of the API: most of the methods you're using have been spitting deprecation warning for over a year by now, and the PyMISP packages released after 2020-01-01 do not support them anymore at all.

I will try to provide a PR (if you're interested), but as I'm not using CAPEv2, I cannot promise the accuracy of the code.

Rafiot:
Note that I didn't test it at all, but it should work fine, and fix #184.


doomedraven:
cool i will ensure the testing, thank you a lot, i had problem with named attributes but it looks like it was much more easier thay it was

PR #192, opened 2020-05-03
Update zip.py

Sankgreall:
- Typo in predefined password: "intefected" rather than "infected"

- If password is not specified in CAPE arguments, password variable is type None, which causes exception on line 48 when we use the .encode() function. Proposed solution is to set the password to blank string if none is specified.

You could also set the value to an empty string when the arguments are created... but I don't know that area of the code!

### DIFF
--- a/analyzer/windows/modules/packages/zip.py
+++ b/analyzer/windows/modules/packages/zip.py
@@ -49,6 +49,17 @@ class Zip(Package):
 
         # Extraction.
         with ZipFile(zip_path, "r") as archive:
+
+            # Check if the archive is encrypted
+            for zip_info in archive.infolist():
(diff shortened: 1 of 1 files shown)


Sankgreall:
Thinking this through, I missed something:

- If the user does not specify a password, this means the archive is not encrypted, the archive is encrypted the the default value (infected), or the archive is encrypted with an unknown password.

- The third option is address through error messaging. To address the first and second points, we need to check to see if the archive is encrypted - if it is and the password is empty, then try using the default password, if not, then just extract.

doomedraven:
Zip package is only should be used if you submit multiple files and one of them is dependency, bcz all the unarchivemwnt is done by my fork of sflock on host before submit to vm

PR #194, opened 2020-05-05
enabled param in community.py wasnt initialized

elmaisbuscado:
Was causing errors like:

enabled.append("feeds")
  File "<stdin>", line 1, in <module>
NameError: name 'enabled' is not defined

or

  File "community.py", line 149, in <module>
    main()
  File "community.py", line 140, in main
    if not enabled:
UnboundLocalError: local variable 'enabled' referenced before assignment


doomedraven:
Gracias

PR #203, opened 2020-05-15
Fix race conditions in Proxmox and our module

michaelweiser:
Hi,

here's the latest changes from my PR against cuckoo (cuckoosandbox/cuckoo#2723), tested and found to be working with current master. Thanks for adding it here! :)

VMs sometimes remain locked for some seconds after a task completed.
They will get stuck in that state if another operation is attempted. So
query the current VM status to extract the lock status to continue
waiting for it to disappear.

Also, our machinery module was carrying internal state which got
corrupted under load. Switch to local variables throughout.


doomedraven:
thanks a lot for update

PR #212, opened 2020-06-10
Using bytes instead of str for syslog data

xme:
With Python3, we need to use bytes instead of str for sock data. Here is the error reported by the module:

2020-06-10 18:52:31,337 [lib.cuckoo.core.plugins] WARNING: The reporting module "Syslog" returned the following error: Failed to send data to syslog server


kevoreilly:
Thank you!

PR #213, opened 2020-06-16
Update filesizes in cuckoo.conf

bartblaze:
More representative this way, 200MB should also still be sufficient.


doomedraven:
did you test `analysis_size_limit = 200000000`? from y experience it can take a lot of time to process huge bsons, like 1-2h, but we have timeout anyway :)

bartblaze:
Yes, should be fine - it was set to 1GB before. Thanks as always! :smile:

PR #233, opened 2020-07-23
Guid randomizer

DayNKnight:
This is my first time doing a pull request, so I am not sure what format to write this in or even if I should have done this on the dev branch. There was a check to make sure that a UUID is never used twice, but that required writing them to a text file, and I was not sure if that was acceptable or not.


doomedraven:
hello, thank you, PR here is fine, well i think it will be hard to overlap with another uuid

PR #241, opened 2020-08-08
updates do_strings in memory.py

zoomequipd:
this resolves some TypeErrors that were generated and steals from procmemory.py to solve it.


doomedraven:
thanks

PR #249, opened 2020-08-14
updated Qakbot parser

threathive:


threathive:
Updated extractor to support parsing either the unpacked parent or the core DLL statically.

kevoreilly:
Awesome!

PR #256, opened 2020-08-26
Update views.py - max-upload error

reynas:


doomedraven:
ah nice catch, thanks

PR #272, opened 2020-09-14
Move license to root project folder

wesinator:
this makes it easier to find the license, github displays the license type info 

- remove old cuckoo docs meta files 
CAPE had versions in root project dir


doomedraven:
thanks

PR #276, opened 2020-09-17
Update supervisor.conf

powerade661:
Suricata no longer uses /var/lib for rules so the supervisor.conf will not use suricata with rules.


doomedraven:
thanks, just small suggestion, i don't recomment to use supervisor, its better to use systemd

PR #283, opened 2020-09-23
Emotet RSA key extraction

phate1:
Hey noticed the RSA key is no longer getting extracted assume they added the extra xor at some point. Put something together that seems to be working my end cant say its the best way to do it but hopefully of some use.
Pete


kevoreilly:
Hi phate1! Thanks a lot for this PR - really nice code. Very happy to see the rsa regex function fixed! I had added yara-based detection for the rsa key too ($ref_rsa) - I have a fix for that which I will push momentarily but this will mean we will have two distinct ways to get the rsa key - strength in depth!

PR #305, opened 2020-10-01
Update procyon java decompiler usage

enzok:
install apt package procyon-decompiler


doomedraven:
thanks for support both versions, as im not sure if they update it so frequently in apt

PR #306, opened 2020-10-01
Add xlm_password option for pwd protected xlm 4.0 macros

enzok:


doomedraven:
i would prefer to ouse the same password argument that we already have instead off adding extra one, just update description no?

enzok:
As long as using "password" in multiple places doesn't break something, I don't see why it can't be used instead of "xlm_password"

doomedraven:
as password argument already says, it also can be used for office files, so to not introduce confusion, changed that and merged

PR #309, opened 2020-10-02
remove duplicate key

enzok:


doomedraven:
also duplicated in 160, removing

PR #322, opened 2020-10-15
Make CAPE work with FOG 1.5.9

hariomenkel:
I edited physical.conf to reflect which settings FOG 1.5.9 needs (especially the API Keys) and edited physical.py accordingly to not rely on HTML parsing anymore but use the API instead. Here is a complete tutorial:

https://mariohenkel.medium.com/using-cape-sandbox-and-fog-to-analyze-malware-on-physical-machines-4dda328d4e2c?source=friends_link&sk=fa7e6be3043697bfcc2decc01ea09443

Tested that it works within a full virtual environment (FOG doesn't care if it images a VM or a physical machine) and also with a physical machine (Intel NUC).

There was also some bugs in multi.py which popped up when using "multi" as machinery.

### DIFF
--- a/conf/physical.conf
+++ b/conf/physical.conf
@@ -2,16 +2,29 @@
 # Specify a comma-separated list of available machines to be used. For each
 # specified ID you have to define a dedicated section containing the details
 # on the respective machine. (E.g. physical1,physical2,physical3)
-machines = physical1
+machines = physical01
+interface = ens33
 
-# Credentials to access the machine
-user = username
-password = password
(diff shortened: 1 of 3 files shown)


doomedraven:
wow thanks for this <33333

can you fix this?

```
Run # stop the build if there are Python syntax errors or undefined names
./modules/machinery/physical.py:293:26: F821 undefined name 'struct'
            send_data += struct.pack(b'B', int(data[i: i + 2], 16))
                         ^
1     F821 undefined name 'struct'

```

PR #324, opened 2020-10-19
Add Hancitor config extractor

threathive:
Decrypts and parses the embedded controllers and campaign/build id from an unpacked Hancitor/Chanitor sample.


kevoreilly:
THANK YOU!

PR #332, opened 2020-10-21
Update Remcos.py

threathive:
Works for newer versions of the RAT, but still has some issues parsing copies which are injected into other processes.


kevoreilly:
You are a legend!

PR #338, opened 2020-10-26
fix machine selection for resubmission and vt download submissions

enzok:


doomedraven:
thanks that helped me to fix the issue, i will pushing that, as i moved this to be just in 1 place instead of check service and in each function

PR #343, opened 2020-10-30
Resolve TypeError in retention reporting module.

r1n9w0rm:
This PR resolves the following generated exception from the retention reporting module:
```

[lib.cuckoo.core.plugins] ERROR: Failed to run the reporting module "Retention":
  File "/opt/CAPEv2/utils/../lib/cuckoo/core/plugins.py", line 783, in process
    current.run(self.results)
  File "/opt/CAPEv2/utils/../modules/reporting/retention.py", line 193, in run
    delete_mongo_data(curtask, lastTask)
  File "/opt/CAPEv2/utils/../modules/reporting/retention.py", line 56, in delete_mongo_data
    if analyses.count > 0:
TypeError: '>' not supported between instances of 'method' and 'int'
```


doomedraven:
oh nice catch, thanks

PR #348, opened 2020-10-31
decode shellcode strings from bytes to strings

sploitem:
I've spotted the problem when there are a shellcode strings, when using API get iocs:
"File "/usr/lib/python3.8/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type bytes is not JSON serializable
"
So, here's a solution.


sploitem:
TYPO

PR #358, opened 2020-11-08
BoxJS Integration tweaks

MatteoManzoni:


doomedraven:
Thanks, but bear in mind rhat continue can be eternal loop that never exit under some weird conditions

Issue #356, opened 2020-11-05
No HTTP/s traffic with InetSIM routing


raw-data:
Noticed few days ago on [CAPE](https://capesandbox.com/) but also true with the latest code from the master repo, that, when processing samples with the InetSIM routing module enabled, HTTP/s connection get somehow lost.

As a reference you can have a look at

1.  https://capesandbox.com/analysis/82106/ win7x86 + InetSIM
2.  https://capesandbox.com/analysis/82107/ win7x86 + TOR

In the first analysis, under `Network Analysis` -> `HTTP` will show 0 entries, in the 2nd one instead, all 7 HTTP GET requests are properly shown.

The behaviour is the same regardless of  x86 or x64 OS, also it's completely independent from the processed sample type (doc, exe, etc.).

Perhaps something "broken" in the rooter module?

doomedraven:
thanks for fix

PR #366, opened 2020-11-18
[chg] handle case where inetsim traffic is not correctly recorded by tcpdump

raw-data:
Probably the original _tcpdump_ filter is fine in the case InetSIM is running on a separated VM (not sure about the noise part ...), nevertheless if it listens on the same _resultserver_ip_ , only DNS traffic is recorded and all the rest (http/https/etc.) is left out.

From here the issues we have with no recorded HTTP sessions (under the _Network_ tab in the report and in the _json_ export) when an analysis was started with InetSIM vs TOR vs Dirty line

### DIFF
--- a/modules/auxiliary/sniffer.py
+++ b/modules/auxiliary/sniffer.py
@@ -141,7 +141,7 @@ class Sniffer(Auxiliary):
         )
 
         # TODO fix this, temp fix to not get all that noise
-        pargs.extend(["and", "not", "(", "dst", "host", resultserver_ip, "and", "src", "host", host, ")"])
+        #pargs.extend(["and", "not", "(", "dst", "host", resultserver_ip, "and", "src", "host", host, ")"])
 
         if remote and bpf:
             pargs.extend(["and", "("] + bpf.split(" ") + [")"])


doomedraven:
1. thanks
2. i strongly not suggest to run inetstim on the resultserver, that doesn't prevent traffic leaks
3. did you test it? it should generate a lot of noise

raw-data:
Good point on the traffic leaks, to be honest I did not take that into account since the whole env is behind a dirty line by its own.

Anyhow, yes I tested it and as stated you can see the differences in the analysis when running InetSim with or without the edited  filter - see below.

InetSim analysis with the original tcpdump filter
inetsim-original-filter

InetSim analysis with the edited tcpdump filter
inetsim-edited-filter

I had no chance to submit the same sample on the community instance, but I reported the same behaviour in #356 for some maldoc samples.

For what concerns the network noise,I did not see any particular stuff (executing the sample via the TOR / Dirty line route shows the same noise anyhow), at least in my case I always see the same background network events someone would expect from Windows OS - but maybe I am missing something you noticed instead.

If removing the whole filter is not "the way to go", we could tight the filtering to get rid of noise, so better support both cases, InetSIM on resultserver_ip vs InetSIM in VM

doomedraven:
thanks

PR #368, opened 2020-11-23
Improving the linux sandbox

idrirap:
I know that linux is a work in progress. I hope I didn't misunderstand the original code.
 
I tried to improve the original documentation and modify the stap module according to the doc instructions. I kept the original .cuckoo folder to avoid breaking existing sandboxes. 

I encountered an issue where the staprun in packages.py would not run because the stap_.ko file was already used by the unstopped subprocess in stap.py. I did not really understand why staprun was run twice. But I fixed this issue by killing the stap.py subprocess after it reached the "systemtap_module_init() returned 0".


doomedraven:
thank you

PR #375, opened 2020-12-03
add helper function and creation script for network_dga_fraunhofer sig

King-Konsto:
This signature detects the likely usage of Domain Generation algorithms by
matching all dns requests of a sandbox run against a pre-built bloomfilter.
The bloomfilter contains the active DGA domains from today, tomorrow and
yesterday and is based on the intelligence of the Fraunhofer DGArchive.

The check against the bloomfilter is really quick. Only if we have hits
we'll proceed and match the DGA domain to its DGA/malware family by
performing a dict lookup that is preloaded in memory.

The bloomfilter and the lookup dict are created by the create_bloom.py
script of the CAPEv2 repository and placed within the /data folder of
the community repository. The update process can be performed
manually via running the script on the local machine before
signature push/pull or via cronjob task.

Relies on PR in the community repo: https://github.com/kevoreilly/community/pull/117

* Bloomfilter library: https://github.com/DCSO/flor
* Fraunhofer DGArchive: https://dgarchive.caad.fkie.fraunhofer.de/welcome/

### DIFF
--- a/cuckoo.pyproj
+++ b/cuckoo.pyproj
@@ -409,6 +409,7 @@
     <Compile Include="lib\cuckoo\common\defines.py" />
     <Compile Include="lib\cuckoo\common\dns.py" />
     <Compile Include="lib\cuckoo\common\exceptions.py" />
+    <Compile Include="lib\cuckoo\common\fraunhofer_helper.py" />
     <Compile Include="lib\cuckoo\common\irc.py" />
     <Compile Include="lib\cuckoo\common\logo.py" />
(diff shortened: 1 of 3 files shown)


King-Konsto:
As there is no requirements.txt and this PR depends on a new library (`flor`) shall I open another PR to add it to the script from https://github.com/doomedraven/Tools/blob/master/Sandbox/cape2.sh ?

doomedraven:
yes we moved dependencies to cape2.sh, but we maybe will partially reintroduce it, i have added like how to install it to simplify life to not skilled people, thank you for your contribution

PR #378, opened 2020-12-07
fixed update capa-rules

ClaudioWayne:
Hi, 
The update of the Capa-Rules via community.py -cr did not work for me. 

1) Errno 39] Directory not empty: '/opt/CAPEv2/data/capa-rules-master' -> '/opt/CAPEv2/data/capa-rules

Screenshot from 2020-12-07 11-37-04

-> I decided to delete the old directorys to replace them with new ones.

2)  Always get the help function although the function was successful

-> return after flare_capa_rules() in the if-statement

There is surely a more beautiful solution, but works for me atm. Feel free to merge if you agree. 
Have a nice day

Claudio

### DIFF
--- a/utils/community.py
+++ b/utils/community.py
@@ -3,6 +3,7 @@
 # This file is part of Cuckoo Sandbox - http://www.cuckoosandbox.org
 # See the file 'docs/LICENSE' for copying permission.
 
+import shutil
 import os
 import sys
 import zipfile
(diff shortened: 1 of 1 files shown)


doomedraven:
good catch, thank you

doomedraven:
about return, not really needed, as that breaks if you put -waf -cr, but the rest is fine

ClaudioWayne:
Yep you are right, looks fine now. 
I have a litte question apart from this topic if you have time. Have you ever been confronted with VirtIO drivers when you set up your Win7 VM with KVM/QEMU? Or is there a way to avoid these VirtIO stuff and install Win 7 in an easy way :). Win 10 works fine out of the box

doomedraven:
em can you elaborate what do you mean? im not sure, if you speak about virtio drivers, you don't need them at all

ClaudioWayne:
i have the following problem when installing win7 with kvm
mouse and keyboard do not working properly either. 

grafik

So i came across this article by a quick research:
https://github.com/hpaluch/hpaluch.github.io/wiki/Install-Windows7-on-KVM-Qemu

but i guess win 7 should work without this VirtIO drivers. I have to investigate the probem more tomorrow. Thanks for your time.

doomedraven:
https://www.doomedraven.com/2020/04/how-to-create-virtual-machine-with-virt.html
pay attention to step 7

doomedraven:
i just checked that link that you posted, that will make vm detected by malware just checking for virtio driver, use my script kvm-qemu.sh and tthis blogpost and thats all :)

PR #390, opened 2020-12-20
Create parallax.yar

r0ny123:
Adding the parallax yara rule from @bartblaze here, it has a match on this [Injected PE Image: 32-bit executable](https://www.capesandbox.com/analysis/108604/#) :)

image

### DIFF
--- a/data/yara/CAPE/parallax.yar
+++ b/data/yara/CAPE/parallax.yar
@@ -1,0 +1,22 @@
+rule Parallax
+{
+meta:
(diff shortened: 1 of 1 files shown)


doomedraven:
hello, thank you, i will merge and move it to https://github.com/kevoreilly/community, please push there next time

r0ny123:
Ok, will do but I don't think it has false positives.

doomedraven:
that isn't due to FP, is due that yaras are done by community

PR #405, opened 2021-01-06
fix requirements.txt syntax

nbargnesi:
Follow-up on 8b910f0f to fix a handful of dependency names in requirements.txt.


kevoreilly:
Thanks!

PR #408, opened 2021-01-10
Deprecate OlympicDestroyer sig due to FP on Cryptbot

r0ny123:
eg. https://capesandbox.com/analysis/111175/


doomedraven:
thanks

doomedraven:
now you need to create account on public sandbox

r0ny123:
> 
> 
> now you need to create account on public sandbox

Yeah already did :) and decision is well taken ;)

doomedraven:
great

PR #415, opened 2021-01-20
Update Zloader.yar

r0ny123:
update for https://www.capesandbox.com/analysis/112957/


kevoreilly:
Thanks a lot for the heads up - Zloader has a config decoder as well as a yara sig so unfortunately to do a proper job this needed updating as well - it is also dependent on the yara sig. I've chosen to loosen the existing $decrypt_config pattern rather than adding another. It's fixed in https://github.com/kevoreilly/CAPEv2/commit/1d1416037834da979899f568cd0864b78a1ba5cb

r0ny123:
Thanks Kevin.

PR #440, opened 2021-02-10
Update report.html

enzok:


doomedraven:
i have removed that as you can see that data is loaded via ajax, and i don't have time to add the check if any memdata exist

Issue #434, opened 2021-02-08
The default timeout for analysis in the cuckoo.conf is not used in the WEB GUI on the Submition Form


anykeypress:
## This is opensource and you getting __free__ support so be friendly!
* Free support from doomedraven ended, no whiskey no support. For something he updated the documentation :)

# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I'm reporting the issue to the correct repository (for multi-repository projects)
- [x] I'm have read all configs with all optional parts

# Expected Behavior
The default timeout in the cuckoo.conf
[timeouts]
default=300
should be present in the Submit form for timeout

# Current Behavior
The value in the Submit form is always 200

# Failure Information (for bugs)

## Steps to Reproduce

1. Set the defaut to anything other than 200
2. Restart CAPE
3. Select "Submit" and observe the timeout field

## Context

| Question         | Answer
|------------------|--------------------
| Git commit       | 783785d3081ad77269334725abf71b3bc8ececfe
| OS version       | Ubuntu 20.04, Windows7 x64 Guest, KVM

## Failure Logs
N/A

As a suggestion (worked for me)
Add the following to the /web/submission/views.py
`enabledconf["timeout"] = cfg.timeouts.default`
underneath
`enabledconf["tlp"] = web_conf.tlp.enabled`

Edit the /web/templates/submission/index.html
Search for
`<input type="text" class="form-control" id="form_timeout" name="timeout"/>`
and replace this with
`<input type="text" class="form-control" id="form_timeout" name="timeout" value="{{config.timeout}}" />`

ClaudioWayne:
Nice thx, works for me.
Consider PR

anykeypress:
@ClaudioWayne , you are welcome to PR it. I'm still in the process of setting my environment up.

PR #447, opened 2021-02-15
Default timeout in cuckoo.conf present in webinterface

ClaudioWayne:
Reference: https://github.com/kevoreilly/CAPEv2/issues/434


doomedraven:
Thanks

PR #472, opened 2021-03-31
update description & cape_type info for IcedIDStage1

r0ny123:


kevoreilly:
I'm afraid cape_type field is not intended for divisions within a malware family, like 'Stage 1' since this is the name that becomes the 'detection' - it was a mistake to allow this historically with IcedId but I will fix this and get back to the place where 'detections' are pure malware family names and nothing else, as that is the design.

kevoreilly:
Since the description is also captured/displayed this is the place for any extra info so it can displayed if useful, but won't affect the overall detection name from being just the family.

doomedraven:
isn't this is the same as photoloader or gziploader? i mean stage 1?

r0ny123:
Yes, in this case it is gziploader.

kevoreilly:
Maybe - I didn't write or merge those. It's tricky to maintain this project to the design standard! I will correct those too.

Consider this for yet another reason these are bad: currently in cape there is only room for one family in detection. Thus if we have stage 1 then stage 2, which do we detect? It would end up saying stage 1 when stage 2 is also in there...

doomedraven:
yes is still a work to do to show all detections, but i have somethign that you will like ;) soon when i will have more time to finish

Issue #238, opened 2020-08-03
Execute all exported functions


bartblaze:
Same issue as [Execute all exported functions](https://github.com/ctxis/CAPE/issues/366). Recap:

> Certain implants that perform DLL sideloading include dummy functions to throw off analysis systems or researchers. For example, in the RedLeaves sample seen here on page **5**:
> https://www.accenture.com/_acnmedia/PDF-76/Accenture-Hogfish-Threat-Analysis.pdf#zoom=50
> 
> Could be an idea to to have an option, for example, "allfunctions", and if the DLL package is triggered, to run all exported functions present (as opposed to the "function" option, which allows you to set a specific function yourself).
> 
> In the Behavior Analysis, the Process Tree would then allow to determine which function is real or being used.

Kev's reply:

> This is a great idea and something I have looked at - I realised it would require pefile to be installed in the guest in order to enumerate the exports. I also considered the likelihood of crashing the process if an exported function was called with bad args, so decided it would be best done with a new process for every export. Other than that it should be straightforward. I will try and get this done, thanks for the suggestion.

If anyone's up for it, would be great! :+1:

doomedraven:
done :)

PR #482, opened 2021-04-21
Updated DLL execution package with the ability to run multiple exports at once

snemes:
This solves #238 too.

An alternative solution that statically enumerates the exported function names (e.g. via pefile) unfortunately will not work, as there are some packed/crypted DLL samples where the decryption code resides in the DLL initialization code, and the real export names won't be visible until the DLL is loaded and the decryption/unpacking code runs.

This solution here is backward compatible with the original DLL package, and it accepts the same package parameters, but it adds support for numeric ranges in the "function=" option, that can be defined in any of the following formats: `1-8`, `1..8`, `#1-8`, `#1..8`, `#1-#8`, `#1..#8` (i.e. two numbers with either a `-` or `..` in between, and each number prefixed with an optional `#`). All these would run the first 8 exported function by their ordinal numbers via a for loop using cmd.exe, running each of them with rundll32.exe (so if any of the exports do not exist, or crashes it won't affect the others).

The defaults setting still defaults to run only the first export (`#1`), but that's easy to change by updating the default value on line https://github.com/kevoreilly/CAPEv2/pull/482/files#diff-06920ac50c46d750409ec79adc8695cf8fd32c992fe61f77251f6a9780cb2d74R21

### DIFF
--- a/analyzer/windows/modules/packages/dll.py
+++ b/analyzer/windows/modules/packages/dll.py
@@ -13,13 +13,13 @@ class Dll(Package):
     """DLL analysis package."""
 
     PATHS = [
-        ("SystemRoot", "system32", "rundll32.exe"),
+        ("SystemRoot", "System32", "rundll32.exe"),
     ]
 
     def start(self, path):
         rundll32 = self.get_path("rundll32.exe")
-        function = self.options.get("function", "#1")
-        arguments = self.options.get("arguments")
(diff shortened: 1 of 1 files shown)


doomedraven:
thank you

kevoreilly:
This is great - thank you Sandor!

bartblaze:
Elegant solution - thank you for the work!

PR #500, opened 2021-05-26
Add Office click-around to trigger macros

enzok:
Thanks Will Metcalf


doomedraven:
nice stuff, thank you both

PR #503, opened 2021-05-28
[fix] pyattck dependency check

raw-data:


doomedraven:
i don't like this, import only if enabled, to remove huge load that it adds to load all jsons

doomedraven:
thanks

PR #505, opened 2021-06-01
Fix yara hit function parameter

enzok:
function is also called from processing.py

### DIFF
--- a/lib/cuckoo/common/cape_utils.py
+++ b/lib/cuckoo/common/cape_utils.py
@@ -164,7 +164,7 @@ def convert(data):
 
 def static_config_parsers(yara_hit, file_data):
     """Process CAPE Yara hits"""
-    cape_name = yara_hit.replace("_", " ")
+    cape_name = yara_hit["name"].replace("_", " ")
     cape_config = dict()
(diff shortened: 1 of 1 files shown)


doomedraven:
where it fails? i would like to see log logs

```
 grep -r static_config_parsers .
./lib/cuckoo/common/cape_utils.py:def static_config_parsers(yara_hit, file_data):
./lib/cuckoo/common/cape_utils.py:            config = static_config_parsers(hit["name"], file_data)
./modules/processing/CAPE.py:                config_tmp = static_config_parsers(cape_name, file_data)
./modules/processing/CAPE.py:            tmp_config = static_config_parsers(hit["name"].replace("_", " "), file_data)
```

doomedraven:
asl as you can see it will break the rest as it expects name already not the hit object

enzok:
strange my CAPE.py has this:       tmp_config = static_config_parsers(hit, file_data)
I'll verify any merges with that file.

enzok:
Found a bad merge. Closing this request.

doomedraven:
ok, thanks anyway ;)

PR #512, opened 2021-06-10
Enable reboot of physical machine after analysis without Fog agent

hariomenkel:
Fog agent on physical machines can't be reached after an analysis since WMI gets restarted during an analysis which does not restart the fog agent. It therefor possibly makes sense to enable the route "execute" in agent.py and directly issue the command

shutdown -r -f -t 0

through modules/machinery/physical.py after an analysis has finished. Code has been tested locally

### DIFF
--- a/agent/agent.py
+++ b/agent/agent.py
@@ -33,6 +33,7 @@ if sys.maxsize > 2 ** 32:
 AGENT_VERSION = "0.11"
 AGENT_FEATURES = [
     "execpy",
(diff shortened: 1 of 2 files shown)


doomedraven:
cool thanks @hariomenkel, the uniq think that would need to check is that if attacker can send command to that port and just kill analysis, there was feature like key that was used for communication in cuckoo-modified, i will need to check

PR #513, opened 2021-06-11
Update agent.py

hariomenkel:
Prevent execution of commands from localhost


doomedraven:
nice thanks, i wasn't thinking in that way, but this is much easier solution

PR #514, opened 2021-06-11
[add] functionality to display user tags from recent files submission…

raw-data:
… tab


doomedraven:
thanks, the uniq thing is that config should be in web.conf as that is something related to webgui

doomedraven:
ok i see lol we need to migrate many options, so one day :D

PR #516, opened 2021-06-15
Add Strrat config extraction

enzok:


kevoreilly:
I am not happy with these mods to cape processing module. What is this special parsing list? This clashes with mods I have which are consistent with a push to remove all malware-specific stuff from this module. If you can explain why this list is necessary (as opposed to straight yara + parser) then I can decide the best way to solve that problem.

doomedraven:
well now that i think, yara is matched so the uniq problem there is the payload/config part in naming that fails as far as i understand no?

enzok:
The issue is that the yara matches in procmemory. The config is in the
original jar. As it's currently structured, the yara match is not tied to
the original submitted file. So, the parser will never run. Also, the
procmemory module runs after CAPE, so no yara to match.

This was the least invasive method to handle these situations.

I'm open to better solutions.

kevoreilly:
The real issue is that cape is not designed to work this way. Procmemdump is not enabled by default, nor is it designed to be. Config extraction is not designed to be performed on procmemdump. If I had wanted that I would have designed it that way 5 years ago. Another issue I have is in seeing ad-hoc mods like this to core components and ultimately core design just for the sake of a single malware family. So the way to go about this is not just to bolt on patchwork code and family-specific lists into a processing module, rather tackle any limitations in the design properly.

The way to do this is to find a way to dump the memory region within the monitor, have it properly labelled using 'cape metadata' and then worked through to the parser using the existing mechanism.

I will be removing these mods to the processing module as well as almost all other family-specific code in my next update.

kevoreilly:
If it can be detected in procmemdump yara then it can be detected, dumped and 'labelled' using yara within capemon. Maybe we need to improve capemon's dumping ability here but something like that is the way forward.

kevoreilly:
... or maybe we could enable config extraction via yara in procmemdump - but that would be done in the procmemory.py processing module. I'm still not keen on this as this means if procmemdump not enabled on submission (i.e. default) the config extraction won't work. That's why I think it's better to have the monitor dump and 'label' it properly the 'cape' way.

enzok:
Having the monitor dump and label would certainly be ideal.

enzok:
When you get time to look at this let me know and I'll help any way I can.

PR #519, opened 2021-06-17
fixed some report stuff

ClaudioWayne:
fixed json link
fixed signature color in html report


doomedraven:
oh that is fixed in bootstrap5 branch but thanks

PR #523, opened 2021-06-22
Update static.py

yevhenprotsenko:
We should add UTF-8 decoding otherwise it will break the report https://github.com/kevoreilly/CAPEv2/blob/841b714d54798293cb1113d10a96a10bcc0f744a/modules/reporting/jsondump.py.


doomedraven:
дякую)

PR #535, opened 2021-07-11
Feature/add signatures reject list

AvihayST:


doomedraven:
nice thanks ;) i will pick by hand lite report as i don't want extra folders having the reports folder already so it can be there as lite.json

AvihayST:
Hi @doomedraven 
Sorry for the messy PR, I chose the wrong fork :[
Glad you like the lite report, promise ill open a PR with that soon :]

doomedraven:
well for me lite report is /iocs/ but always when the stuff can be on/off im fine to add that

doomedraven:
@AvihayST check how i did lite report https://github.com/kevoreilly/CAPEv2/pull/536
it will store it in reports as lite.json but i have added link to webgui and dl

doomedraven:
i have picked most interesting stuff/fixes from this pr, you can see them where https://github.com/kevoreilly/CAPEv2/pull/537

btw you introducing new dependency `netinterfaces`, we already have `psutils` so i have changed that to not add extra dependencies when other libraries can handle that in the same way

PR #543, opened 2021-07-22
Exit cleanly from rooter

gerhard4:
The rooter service doesn't exit cleanly after a `SIGTERM` because of the exception raised in the `recvfrom` exception handler. This change makes it exit cleanly when it should.

Also a couple minor improvements to the rooter logging.


doomedraven:
thank you

PR #548, opened 2021-07-28
Post-Analysis Lookup column was not being populated.

rodac5:
Hello. I found a bug which was preventing the Post-Analysis Lookup column from being populated. The change below fixed the bug.

### DIFF
--- a/web/analysis/views.py
+++ b/web/analysis/views.py
@@ -1092,7 +1092,7 @@ def report(request, task_id):
         )
         network_report = results_db.analysis.find_one(
             {"info.id": int(task_id)},
-            {"network.domainlookups": 1, "network.iplookups": 1, "network.dns": 1, "network.hosts": 1},
+            {"network.domains":1, "network.dns": 1, "network.hosts": 1},
             sort=[("_id", pymongo.DESCENDING)],
         )
 


doomedraven:
Hello thanks but that would break other parts https://github.com/kevoreilly/CAPEv2/search?q=domainlookups

where it wouldn't be populated exactly? i can't see it https://github.com/kevoreilly/CAPEv2/search?l=HTML&q=domains

rodac5:
Hi @doomedraven,
it is not being populated here:
web/templates/analysis/network/_dns_not_ajax.html
and possibly here as well:
web/templates/analysis/network/_dns.html.

I am using inetsim, and my dns queries are returning 127.0.0.1, as configured. However, the "post-analysis lookup" column was always empty, even considering the report.json comprised the real IP under network.domains.

I noticed that the "domains" key is not being retrieved from mongo here:

network_report = results_db.analysis.find_one(
            {"info.id": int(task_id)},
            {"network.domainlookups": 1, "network.iplookups": 1, "network.dns": 1, "network.hosts": 1},
            sort=[("_id", pymongo.DESCENDING)],
        )
And I noticed that by adding "network.hosts": 1 to it solved the "problem" (considering this is really a problem and not an expected behaviour).

Regarding breaking other parts, you are right, I should have proposed the fix below.
{"network.domainlookups": 1, "network.iplookups": 1, "network.dns": 1, "network.hosts": 1, "network.domains":1},

I am wondering now if the post-analysis column not being populated is related to the fact that the response for the bank dns query is empty (instead of 127.0.0.1), and why it is empty.... 
https://pasteboard.co/KdqfWz2.png

doomedraven:
is super strange, i just checked `web/templates/analysis/network/_dns_not_ajax.html` and there is no usage of network.domains at all hm, is why i have posted the search query, there is no `domains` at all 

what im not sure how it works here then? https://capesandbox.com/analysis/175050/

about inetsim and empty field, probably there was some problem and there is no handshake i saw that in past

rodac5:
Hi @doomedraven,
i am not sure it is working here (https://capesandbox.com/analysis/175050/), as the post-analysis column is empty. Maybe because this analysis did not use inetsim, so it did not make sense to run a post-analysis lookup. But I might be wrong, I started using Cape just recently. By the way, it is a great tool!!!

Regarding the pull request, here is my rationale. Sorry if there is any mistake to it:

The values displayed on the post-analysis lookup column from `web/templates/analysis/network/_dns_not_ajax.html`are populated from the dict `domainlookups`:
'{% if domainlookups and domainlookups|get_item:p.request %}
                        {{ domainlookups|get_item:p.request }}'

The keys from such dict are the domains themselves from the ["network"]["domains"] section, and the values are the resolved ips. That is why there are no references for `domains`in that file.
`domainlookups = dict((i["domain"], i["ip"]) for i in network_report["network"]["domains"])`

The thing is the variable `network_report` does not retrieve the ["network"]["domains"] section from mongo.

       ```
 network_report = results_db.analysis.find_one(
            {"info.id": int(task_id)},
            {"network.domainlookups": 1, "network.iplookups": 1, "network.dns": 1, "network.hosts": 1}
            sort=[("_id", pymongo.DESCENDING)],
```
        
I have not dug deeper to find out if  ["network"]["domains"]  is the place where the post-analysis lookup should really be. I simply did a grep for the ip to find the key storing it. So, when I added "network.domains": 1 to it, the post-analysis lookup column started being populated.
         `  {"network.domainlookups": 1, "network.iplookups": 1, "network.dns": 1, "network.hosts": 1, "network.domains":1}`

doomedraven:
ok yes thanks, i just checked in mongo and there is no domainlookups, weird that in code there is kinda also no domains, but now should be fine, im gonna test with merge to ensure that it works just fine

PR #550, opened 2021-07-29
Precisions on agent placement on Windows machines

DavidResin:
I see many people online having issues with the agent because they do not know that the user startup folder will not work. Here is my attempt at clarifying that. Feel free to alter if needed, but this absolutely needs to be in the documentation. I kept the end of the paragraph as-is but I think this should be further improved. Newbie users will have trouble understanding what they have to do with an "etc".

### DIFF
--- a/docs/book/src/installation/guest/agent.rst
+++ b/docs/book/src/installation/guest/agent.rst
@@ -17,5 +17,4 @@ On Windows simply launching the script will also spawn a Python window, if
 you want to hide it you can rename the file from *agent.py* to **agent.pyw**
 which will prevent the window from spawning.
 
-If you want the script to be launched at Windows' boot, just place the file in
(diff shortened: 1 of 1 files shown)


doomedraven:
Thanks David, anything that you consider that can be clarified or improved is more than welcome

 i love this part :) 
```
etc see how malware does it for tricks ;).
```

DavidResin:
You're very welcome. I am struggling a lot to get CAPEv2 to work and it took me a lot of time to figure out little details like these so I really want others to avoid going through the same. Nevertheless, please don't hesitate to be very critical of my suggested edits ; I am by no means an expert in the field of VMs and the last thing I want is to introduce incorrect information in your documentation :)

doomedraven:
Feel free to ask any question, that isnt unlear, or you have doubts im
happy to help to person that trying also to get stuff working on their own,
and not just coming to any project with hey this shit doesnt work 🤣

PR #553, opened 2021-08-05
Disable PredatorPain without log errors

gerhard4:
With the ".py" extension on the file, it didn't get correctly loaded and caused errors in the log.

Removing the extension makes so that it isn't loaded and doesn't cause errors.


doomedraven:
so strange, i had move that to .py_disabled, not sure how .py was added again, thanks

PR #560, opened 2021-08-17
settings.py: reload secret_key import fixed

MiS27:
Import for secret_key reload should be relative - same as initial import.
Bug prevents running some `python3 manage.py` commands on fresh install.
Currently first time command is executed it creates secret_key.py if missing but fails with (re)loading it (due to wrong import). After first failure next commands work fine (file already exist imported with first attempt).


doomedraven:
thank you

PR #566, opened 2021-08-24
Deadlock in cuckoo.core.rooter module has been fixed

a1ext:
Deadlock in `cuckoo.core.rooter` module has been fixed.
There were cases when the cape is frozen when all the threads were waiting on the `lock.acquire()` call (but before some code raised an exception in between the lock acquire and release so the `lock` remained locked.
This PR fixes the issue so even if the code will raise an error, the lock will be released anyway.

### DIFF
--- a/lib/cuckoo/core/rooter.py
+++ b/lib/cuckoo/core/rooter.py
@@ -60,32 +60,29 @@ def rooter(command, *args, **kwargs):
         log.critical("Unable to passthrough root command (%s) as the rooter " "unix socket doesn't exist.", command)
         return
 
-    lock.acquire()
+    ret = None
+    with lock:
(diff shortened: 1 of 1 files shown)


doomedraven:
hello @a1ext, interesting solution, i was looking to fix that in past but i was focused more on scheduler.py, so thank you a lot for your time and effort on this. this is huge help for us, and you killed an entry in my ToDo what was sitting for very long time

a1ext:
> hello @a1ext, interesting solution, i was looking to fix that in past but i was focused more on scheduler.py, so thank you a lot for your time and effort on this. this is huge help for us, and you killed an entry in my ToDo what was sitting for very long time

Hi @doomedraven , you are welcome! 😄

PR #568, opened 2021-09-01
Update urls.py

yevhenprotsenko:


doomedraven:
hello, could you explain why? it works for me with that, and it allows remove pending tasks, so if i merge it breaks another thing, so i need some context why do you mod this, thanks

yevhenprotsenko:
hello, the issue I cannot delete the specific task from CAPE using `Delete Task` API node.

api

doomedraven:
thanks i have updated your PR,  when i tried i tried via web, now tested and you are right, works now, thanks for headups

PR #571, opened 2021-09-08
Fix PE and .NET file infos

enzok:


doomedraven:
thanks, just one small note, please don't use exception without show it, as in past it gave me a lot of headache trying to make some code works

PR #572, opened 2021-09-10
Refactor malware family detection

gerhard4:
For some changes in our fork I wanted the malware detection before the processing stats handling. To me, this makes sense as the malware detection is sort of a processing step, and the stats and error log handling coming after all processing steps makes sense to me.

While doing this, I factored the malware detection out into a separate function, cleaned the code up a bit, and maybe even fixed a bug or two. It seems it hasn't worked as intended in all cases (see for example the use of the `not family` condition in `elif` clauses).

But since I'm not fully aware of how all the data is produced that is used here, please review the changes.

### DIFF
--- a/lib/cuckoo/core/plugins.py
+++ b/lib/cuckoo/core/plugins.py
@@ -393,6 +393,8 @@ class RunProcessing(object):
         else:
             log.info("No processing modules loaded")
 
+        self._detect_family()
+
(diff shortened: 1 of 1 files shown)


doomedraven:
looks good to me, thanks

kevoreilly:
Thanks @gerhard4 !

PR #573, opened 2021-09-10
Rename option `logging` to `log_rotation`

gerhard4:
The current name feels confusing, as the options in this section don't affect logging itself, but rather the log rotation.


doomedraven:
thank you

PR #595, opened 2021-10-06
Fix on-demand functionality for dropped files

psalire:
Hi, something to fix clicking "on-demand" buttons for dropped files currently resulting in "File not found" errors.
This is b/c it was looking for dropped files in a directory called `dropped` (which doesn't exist) rather than in the `files` directory. There was also no code for saving on-demand dropped file results, which I've addressed


doomedraven:
thank you

PR #620, opened 2021-11-05
Add arch key to machine config and fix task fetching

enzok:
requires 'alembic upgrade head'
requires 'arch' key added to machinery file


doomedraven:
thank you, had posibility to test

PR #634, opened 2021-12-05
feature(dumping): add support for receiving ppid from the dumper

davidt99:
This will allow getting the parent process id in the files.json file in the reports. This is coupled to capemon PR (https://github.com/kevoreilly/capemon/pull/33).
This should be the first PR of series of PRs aiming to align our fork with the mainline.


kevoreilly:
Thank you

PR #637, opened 2021-12-07
black -l 144

TheMythologist:
Ran command `black -l 144 .` to make life easier for me when reformatting code :)


doomedraven:
my devops forces me to uses -l 132, so i can't get that in, but between 132 and 144 there is no much diff

TheMythologist:
Sure I'll edit accordingly :)

doomedraven:
about re2, if you just replace without test, please don't do that, as we had many situations where re2 was complaining and breaking, the exceptio nad none part is nice, im checking how to cherry pick it, but is better to provide separated PRs for each block so that simplifies a lot merging

doomedraven:
i have cherry-picked removing noen and add Exception, thank you

PR #638, opened 2021-12-07
Update hashes for pyxlsb2 and xlrd2.

tbeadle:
pyxlsb2 and xlrd2 have had commits on master and so the hashes for
master.zip of those repos has changed. This is a band-aid for a problem
in the way XLMMacroDeobfuscator defines in dependencies. I have opened
https://github.com/DissectMalware/XLMMacroDeobfuscator/issues/100 to
hopefully get the author of that project to address this in a way that
isn't so brittle.


kevoreilly:
Thanks

PR #643, opened 2021-12-12
fix(web/templates/apiv2): fix `config.tasks_latest` typo

nikAizuddin:
Looks like a typo in `web/templates/apiv2/index.html`:
https://github.com/kevoreilly/CAPEv2/blob/4da21045aa1f87bfbfdf3fef3351d787fa1b5568/web/templates/apiv2/index.html#L597-L613

I can't find any `[latests]` config in https://github.com/kevoreilly/CAPEv2/blob/4da21045aa1f87bfbfdf3fef3351d787fa1b5568/conf/api.conf. There is only `[tasks_latest]`


doomedraven:
thanks

Issue #644, opened 2021-12-12
Termination ends prematurely if systemtime is modified


TheMythologist:
Current code:
https://github.com/kevoreilly/CAPEv2/blob/4da21045aa1f87bfbfdf3fef3351d787fa1b5568/analyzer/windows/analyzer.py#L543-L553

In most situations this code is fine, but in certain samples where the system time is modified, the execution may terminate prematurely.

E.g. https://capesandbox.com/analysis/213395/
Python script that was run:
```python
import datetime
import win32api

now = datetime.datetime.now()
print (now)
newday = now - datetime.timedelta(days = 8)
print (newday)

year = int(newday.year)
month = int(newday.month)
# Return the day of the week as an integer, where Monday is 0 and Sunday is 6.
dayOfWeek = int(newday.weekday())
day = int(newday.day)
hour = int(newday.hour)
minute = int(newday.minute)
second = int(newday.second)
millseconds = newday.microsecond // 1000

win32api.SetSystemTime(year, month, dayOfWeek, day, hour, minute, second, millseconds)

i = 0
while True:
    with open(str(i) + ".txt","w") as f:
        f.write("hello test")
```

The execution will also terminate prematurely if system time is increased by the timeout duration (e.g. if the timeout is 200 seconds, changing the time from 8:10 to 8:15 will end the execution)

I'll try to figure out a way to measure analyzer runtime duration without using datetime. Feel free to throw any ideas below :)

kevoreilly:
Why is the system time being modified? I think this is the root of your problem. I'm not sure this is something we should be working around...

TheMythologist:
I agree that most malware samples would not modify system time (simply because doing so would immediately raise some red flags), but I think it would be better to address this issue since after all this is a "sandbox", meaning we should be able to throw almost anything and let it be able to run successfully and to completion.

This issue could take a backseat for now.

kevoreilly:
I disagree. If you have a sample that does this I'd be interested to see it. I've been in this game for many years and never seen such a thing.

kevoreilly:
This particular issue, were it to be seen in actual malware, would be best countered by a hook in the monitor for the SetSystemTime API. There are many existing hooks that prevent malware doing things that are undesirable in the sandbox.

[...]

PR #645, opened 2021-12-13
Change from datetime to timeit

TheMythologist:
Fix for #644 

Changes from `datetime.now()` to `timeit.default_timer()`, which is not dependent on the system-wide clock.

It's a simple fix, thus I chose to include it.


kevoreilly:
Thanks for the PR - timeit does seem to be better so I'll merge once I've had a chance to test.

doomedraven:
i did test it, looks like everything works just fine

PR #646, opened 2021-12-13
drarft: Various fixes

davidt99:
Few fixes:
- Performance improvement for querying cape task status
- Cleanup mongo by default. This is breaking changes in a way since the flag is changed now to `--dont-delete-mongo`
- Handle network error in the scheduler
- Remove FP of process hollowing for acrobat reader

PR #647, opened 2021-12-14
Cleaned up analyzer folder

TheMythologist:
- Remove all full-stops at end of log statements for standardisation
- Fix a few typos
- Miscellaneous improvements to prevent code duplication
- Convert to f-strings for better performance
- Remove all unnecessary default arguments
- Use `with` operator for better exception handling
- Reorder import statements and remove unused imports

This is quite a big pull request but hopefully it's fine, let me know if you want anything changed / you don't agree with any standardisation and I'll change accordingly


doomedraven:
I would suggest to split those changes in different PRs, that would simplify review and merge faster as i see some removal of comments that i would like to keep and many other small things. in general this a nice code update, but will take tons to review and merge

doomedraven:
ok few things>

1. did you test your changes?
2.  i see changes like next, and they will break in some cases as we don't have time to investigate from where str/byte diff is send, and i see al ot of small changes like those removed, i guess they are done with another regexes or something like that?
```
                -pids.encode() if isinstance(pids, str) else pids,
                -metadata.encode() if isinstance(metadata, str) else metadata,
                -category.encode() if isinstance(category, str) else category,
                +pids.encode(),
                +metadata.encode(),
                +category.encode(),
```


so again i do appreciate your effort to update cape code, but blind regex replace isn't the best, for string like `%s`to `f-strings` i guess is ok, but for many others parts is NOT ok, so as i told is best to split different PRs to be able to review them separately and merge what is ok

TheMythologist:
Hi, by "hopefully this is ok" I actually mean that yall wouldn't mind the massive PR, but i understand that splitting into small PRs will work better.

I have actually manually checked each and every single file and typed these changes - these commits were done without using any regex replaces. As for the reason why the `isinstance` was removed, I'm assuming that the user input is of the correct format (`str`), as indicated in the comment above that all arguments should be string. Let me know if you would like to keep the type-checking in place (which I personally disagree with).

Most of the comments that I've removed are actually duplicates of code that are already present.

I will go ahead with creating smaller PRs for future changes, should I close this and split this PR into smaller changes? With these smaller PRs I can also include comments in changes that may be controversial to hopefully explain my rationale of the changes.

TheMythologist:
Note that in changes where I removed the type-checking, I have made sure that all current references/uses of the function are of the correct type and do not break any of the changes. I disagree with type-checking because it is simply impossible to blanket-cover all possible types that a user may accidentally use instead of the correct type. I am thinking of adding Python type-hinting in future changes, should I merge such changes only then?

doomedraven:
well even hand changing instead of regex, sound like you didn't run that in production/dev? so as i told yes it should be str but sometime is bytes and is a mess but i don-t have time to investigate from where it comes to be able to get the rid of the root problem.

Python hinting is nice, but you know the project is massive and a lot of code is done by contributors too and is super hard to keep that all of the same type. is why we have those ugly hacks

the problem of this massive PR is to test it, as you know there can be many pitfails in many different parts, as example that bytes vs strings. So i do understand that is lame to split PR into small parts when you on fire, but we doing the project in our free time and we don-t have a lot of free time to get so massive PR without proper review(it can be ok or can be not) and then just wait for people to complain if they will spot something wrong before of us

doomedraven:
also what is the goal to do changes like those? you just rebuild the same string

Screenshot 2021-12-15 at 09 48 36

doomedraven:
so for example see how small PR as this even contains bugs https://github.com/kevoreilly/CAPEv2/pull/646 so imagine how hard is to review and test big one

ethhart:
While I agree readability is improved with f-strings, it is not optimal for the python logging module as each string is formatted regardless of the specified logging level. This especially adds up when applied to all the debug statements. It also doesn't work as well with log aggregators. Seeing as this PR is meant to address performance improvements and code cleaning, it doesn't make sense to move backwards.

TheMythologist:
Yup I've taken note of the logging performance (f-string performance is better but logging takes care of its optimizations), I'll edit accordingly.

doomedraven:
so idk how to follow here guys, is nice cleanup, but i can't merge that directly, and im almost about to go 1 by 1 file review to push them separately to do proper review %)

if anyone have a better advice let me know, but i see parts that i would like restore, we just specify maybe with comment when few small things was done in different way, specially separated pillows imports

TheMythologist:
Don't worry I'll split the PR up into smaller functionalities (e.g. f-string conversion, following PEP-8 conventions, etc.). No actions needed for now, I'll close the PR and submit each change as each PR.

doomedraven:
Screenshot 2021-12-17 at 15 45 17
ok few notes, apart of already is instance that i told before, keep this as it is please, due to what we saw in past sometime import of pillow can take 5 or more minutes is why we have that with logging in that way

PR #648, opened 2021-12-15
fix copy_path delete bug

N1neSun:
when cuckoo config delete_bin_copy is on, using static,url,pcap may cause the bug, process will failed.because copy_path is None.


doomedraven:
thank you

PR #652, opened 2021-12-19
Use default parameter for `.encode()` and `.decode()`

TheMythologist:
Changed `.decode("utf-8")` and `.encode("utf-8")` to `.decode()` and `.encode()` respectively
Note: `.decode("utf-8")`, `.decode("utf8")` and `.decode()` all do the same thing

May seem like a small update but I think it goes a long way in improving code readability


doomedraven:
thank you

PR #653, opened 2021-12-19
Sort imports for analyzer, admin and agent folders

TheMythologist:
- Added pyproject.toml to standardise isort configuration
- Removed unused imports

### DIFF
--- a/analyzer/linux/lib/api/screenshot.py
+++ b/analyzer/linux/lib/api/screenshot.py
@@ -1,24 +1,14 @@
 # Copyright (C) 2010-2015 Cuckoo Foundation.
 # This file is part of Cuckoo Sandbox - http://www.cuckoosandbox.org
 # See the file 'docs/LICENSE' for copying permission.
+
 from __future__ import absolute_import
 import logging
-
-log = logging.getLogger(__name__)
-
-log.debug("Importing 'math'")
 import math
-import logging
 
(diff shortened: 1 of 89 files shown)


analyzer/linux/lib/api/screenshot.py
```diff
@@ -1,24 +1,14 @@
...
 
 log = logging.getLogger(__name__)
 
 try:
-    log.debug("Importing 'pyscreenshot'")
-    try:
-        import pyscreenshot as ImageGrab
-    except ImportError:
-        log.error("Missed dependency: pip3 install pyscreenshot")
-        HAVE_PIL = False
```
TheMythologist, line 21: `pyscreenshot` was removed as ImageGrab would simply be overwritten by `from PIL import ImageGrab`.
In cases where `from PIL import ImageGrab` fails, `HAVE_PIL` would be False and `ImageGrab` would not be called, so there is no difference

TheMythologist:
Feel free to edit `pyproject.toml` as you see fit

analyzer/linux/lib/api/screenshot.py
doomedraven: ok thanks for clarification

doomedraven:
thanks wanted to add `isort` for long time but never had time to start with

PR #654, opened 2021-12-21
Cleanup analyzer folder

TheMythologist:
I apologise but this is a relatively big PR as I have no idea how to split it up into smaller chunks.

Do leave a comment if you disagree with any of the changes and I'll explain my rationale for the changes.

Here is a main summary of the changes:

1. Use context manager `with`
2. Change `.split()[-1]` to `.rsplit(maxsplit=1)[-1]`
3. Comparison `is None` instead of `== None` (similarly for `True` and `False`)
4. Remove unnecessary lines/variables


analyzer/linux/lib/common/results.py
```diff
@@ -108,15 +106,13 @@ def init(self, dump_path, filepath=False, pids="", metadata="", category="files"
         """
         All arguments should be strings
         """
-        if pids:
+        if pids and not isinstance(pids, str):
```
doomedraven, line 109: Not sure if this is correct

doomedraven:
looks good so far, thanks

doomedraven:
there is bug, fixing
```
2021-12-21 20:15:18,153 [lib.common.results] ERROR: Exception uploading file C:\dWVIkvNK\CAPE\2520_165082518353223122021 to host: 'list' object has no attribute 'encode'
  File "C:\tmpp7oey4gv\lib\common\results.py", line 34, in upload_to_host
    nc.init(dump_path, file_path, pids, ppids, metadata, category, duplicated)
  File "C:\tmpp7oey4gv\lib\common\results.py", line 123, in init
    ppids.encode(),
AttributeError: 'list' object has no attribute 'encode'
```

doomedraven:
fixed there was a typo with ppids vs pids, thanks for code cleanup

doomedraven:
i have restored thgis part

```
if pids and not isinstance(pids, str):
            pids = " ".join(pids)
        else:
            pids = ""
        if ppids and not isinstance(ppids, str):
            ppids = " ".join(ppids)
        else:
            ppids = ""
```

as i found more problems, i have added logging to see what was there and it is empty, empty list or list with int pid, the best would be to go and properly fix that but i don't have time to investigate now

doomedraven:
ok removed those elses, found problematic part

TheMythologist:
Thanks for the merge, in the future I hope to add type hinting so we don't have to use these ugly hacks, is that ok?

doomedraven:
yes that would be amazing, is super useful but you know start refactoring such a big piece of code is hard when you also want to add new features

doomedraven:
thanks for all the code cleanup/improvements

PR #655, opened 2021-12-21
Cleanup of analyzer packages

TheMythologist:
Summary of changes:

1. Remove unnecessary `from __future__ import absolute_import`
2. Move adding of file_extensions to `common.py` to reduce code duplication


doomedraven:
this is really nice cleanup :) thank you

PR #656, opened 2021-12-22
Update analyzer.py

TheMythologist:
The `or` segment of line 762 was removed as it was actually unnecessary.

There are 3 scenarios:
**First scenario**:
`self.files[filepath.lower()]` exists as a string. This would be the ideal scenario.

**Second scenario**:
`self.files[filepath.lower()]` doesn't exist, this would return `""` due to the `dict.get()`

**Third scenario**:
`self.files[filepath.lower()]` exists but is not a string. The code would have failed above already because `filepath.lower()` was called earlier.

### DIFF
--- a/analyzer/windows/analyzer.py
+++ b/analyzer/windows/analyzer.py
@@ -758,9 +758,8 @@ class Files(object):
             file_details = self.files_orig.get(filepath.lower())
             category = file_details["category"]
             metadata = file_details["metadata"]
-            # Do not remove or, as if no it will return empty list instead of string
-            pids = self.files.get(filepath.lower(), "") or ""
-            filepath = self.files_orig.get(filepath.lower(), {}).get("path", filepath)
+            pids = self.files.get(filepath.lower(), "")
+            filepath = file_details.get("path", filepath)
 
         if category == "memory":
             if pids:
--- a/analyzer/windows/analyzer.py
+++ b/analyzer/windows/analyzer.py
@@ -758,7 +758,7 @@ class Files(object):
             file_details = self.files_orig.get(filepath.lower())
             category = file_details["category"]
             metadata = file_details["metadata"]
-            pids = self.files.get(filepath.lower(), "")
+            pids = self.files.get(filepath.lower(), [])
             filepath = file_details.get("path", filepath)
 
         if category == "memory":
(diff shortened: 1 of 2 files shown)


doomedraven:
without `or` it still returns `[]` so that requires more in depth fixing so i will leave this open till that will be fixed

doomedraven:
i think the main problem is here 
`self.files[filepath.lower()] = []`

TheMythologist:
Ahh my apologise I see when I went wrong now, I'll push a commit soon hopefully to fix this issue

doomedraven:
thanks for help

TheMythologist:
I think we should leave it as a list, and no errors will be thrown in the execution chain that continues after this line of code. `pids` was always meant to be a list of pids (as can be seen from `pids[0]`, so I think we should leave it as it is. It will be nicely converted into a string in `nc.init()`.

doomedraven:
yes and no 
as some files are captured without pids, the pids variable will be `[]` that is `False` so that won't enter the if pids do join

doomedraven:
here is the example from yesterdays testing

```
2021-12-21 22:21:58,581 [lib.common.results] INFO: [] - pid - <class 'list'>
2021-12-21 22:21:58,596 [lib.common.results] INFO:  - ppid - <class 'str'>
2021-12-21 22:21:58,596 [lib.common.results] ERROR: Exception uploading file C:\ProgramData\Microsoft\Windows\DRM\drmstore.hds to host: 'list' object has no attribute 'encode'
  File "C:\tmpp7oey4gv\lib\common\results.py", line 34, in upload_to_host
    nc.init(dump_path, file_path, pids, ppids, metadata, category, duplicated)
  File "C:\tmpp7oey4gv\lib\common\results.py", line 124, in init
    pids.encode(),
AttributeError: 'list' object has no attribute 'encode'
```

TheMythologist:
Not sure if this is the most optimal way but I think this is the best way currently because there is no reason to check if `pids` is True or False. Will have to revisit this when doing future refactoring so that we don't have to use this ugly hack.

doomedraven:
So that breaks again, i will keep that as it was before

```
2021-12-23 10:07:54,468 [lib.common.results] ERROR: Exception uploading file C:\curtain.log to host: can only join an iterable
  File "C:\tmpegtirc9m\lib\common\results.py", line 34, in upload_to_host
    nc.init(dump_path, file_path, pids, ppids, metadata, category, duplicated)
  File "C:\tmpegtirc9m\lib\common\results.py", line 115, in init
    pids = " ".join(pids)
TypeError: can only join an iterable
```

doomedraven:
it will require more detailed review of upload func to fix that properly but in 2022 :)

PR #657, opened 2021-12-22
Ran isort on all files

TheMythologist:
Only ran one command to generate all these changes:
`isort .`
No other changes were made, so there should not be any changes to the logic/functionality of the files.
Ran according to the configuration set in `pyproject.toml`


TheMythologist:
Hi, could you rerun isort with the specified configuration options in `pyproject.toml`? I think a few imports were messed up by the `python-package.yml` (I think you'll need to update that too)

For instance in `admin.py`:
```
from admin_conf import (JUMP_BOX, VOL_PATH, CAPE_PATH, MASTER_NODE, CAPE_DIST_URL, JUMP_BOX_PORT, JUMP_BOX_USERNAME,
                        REMOTE_SERVER_USER, SERVERS_STATIC_LIST)
```

^ The imports are no longer alphabetically sorted.

doomedraven:
ok so we can't use isort, as it breaks some of the 3rd part integrations like capa for example that breaks the rest

PR #658, opened 2021-12-22
fix many same samples delete bugs

N1neSun:
when cuckoo config delete_bin_copy is on，upload multiple same samples may cause the bug,because sample may be deleted，Later analysis can not be completed.


doomedraven:
oh nice thank you, now i can remove this from ToDo :)

PR #660, opened 2021-12-23
Logging and print cleanup for lib/cuckoo/core

TheMythologist:
Cleanup for files in lib/cuckoo/core

1. Use f-strings for non-logging strings
2. Use % format for logging strings
3. Standardisation - Remove last full-stops for logging and printing strings

### DIFF
--- a/lib/cuckoo/core/database.py
+++ b/lib/cuckoo/core/database.py
@@ -598,9 +598,7 @@ class Database(object, metaclass=Singleton):
                 self.engine = create_engine(connection_string)
         except ImportError as e:
             lib = e.message.split()[-1]
-            raise CuckooDependencyError(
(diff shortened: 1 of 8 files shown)


doomedraven:
thank you

doomedraven:
so issue was here

```
{"name": current.class.name, "time": float(f"{timediff.seconds}.{timediff.microseconds // 1000: 03d}")}
should be
{"name": current.class.name, "time": float(f"{timediff.seconds}.{timediff.microseconds // 1000:03d}")}
```

doomedraven:
ok there is something else wrong inside of the vm logging i guess, i will keep that to fix when i will have time and not smash my PTO, so i will slow dawn merges and code review, but i appreciate all your job

doomedraven:
ok found the problem where, it has missed `f` for config generation fixed and now all works

PR #675, opened 2021-12-27
Reformat imports for modules folder

TheMythologist:
- Removed unused imports


doomedraven:
thank you

PR #676, opened 2021-12-28
Reformat imports for lib folder

TheMythologist:
- Removed unused imports


doomedraven:
thanks

doomedraven:
and special thanks for pushing them in small portions, that makes it easier to test and rewiew

PR #677, opened 2021-12-28
Fix variable name

enzok:
All these file reformats should have had some testing


doomedraven:
Thanks, well tests, nobodu wants to write them 🤣

PR #678, opened 2021-12-28
Range default parameters

TheMythologist:
- Change `range(0, x)` to `range(x)`
- Change `range(len(...` to `enumerate(`

### DIFF
--- a/modules/processing/memory.py
+++ b/modules/processing/memory.py
@@ -62,8 +62,7 @@ class ReturnJsonRenderer(JsonRenderer):
             # Nodes always have a path value, giving them a path_depth of at least 1, we use max just in case
             acc_map, final_tree = accumulator
             node_dict = {}
-            for column_index in range(len(grid.columns)):
-                column = grid.columns[column_index]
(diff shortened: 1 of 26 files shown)


modules/processing/memory.py
```diff
@@ -62,8 +62,7 @@ def visitor(
             # Nodes always have a path value, giving them a path_depth of at least 1, we use max just in case
             acc_map, final_tree = accumulator
             node_dict = {}
-            for column_index in range(len(grid.columns)):
-                column = grid.columns[column_index]
+            for column, column_index in enumerate(grid.columns):
```
doomedraven, line 65: personal comment to review if index is properly used

```
>>> data = [1,2,3,4]
>>> for index, data in enumerate(data):
...    print(index, data)
...
0 1
1 2
2 3
3 4
```

doomedraven:
thanks, i have fixed small fix in memory.py

doomedraven:
Btw you might be interested in submit some of this fixes to the authors of the tools like Didier's, kixtart to make it more optimized in main repo

PR #680, opened 2021-12-28
Set maxsplit in `.split` and `.rsplit`

TheMythologist:
Small improvement in code by setting limit on `maxsplit`.
`.split` was changed to `.rsplit` in cases where only the last element was needed.

Unlike the other PRs, I have not individually reviewed these changes yet (as it's 1am in my timezone right now lol), I'll comment again when I've reviewed the changes myself (or if you want to you can review them too :D)

### DIFF
--- a/lib/cuckoo/common/abstracts.py
+++ b/lib/cuckoo/common/abstracts.py
@@ -690,7 +690,7 @@ class Processing(object):
         self.shots_path = os.path.join(self.analysis_path, "shots")
         self.pcap_path = os.path.join(self.analysis_path, "dump.pcap")
         self.pmemory_path = os.path.join(self.analysis_path, "memory")
-        self.memory_path = get_memdump_path(analysis_path.split("/")[-1])
+        self.memory_path = get_memdump_path(analysis_path.rsplit("/", 1)[-1])
         # self.memory_path = os.path.join(self.analysis_path, "memory.dmp")
         self.network_path = os.path.join(self.analysis_path, "network")
         self.tlsmaster_path = os.path.join(self.analysis_path, "tlsmaster.txt")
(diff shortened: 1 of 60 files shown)


doomedraven:
it looks good, but maybe that is just confusion, you added `maxsplit`, then removed anothers, and on the rest you don't use it

TheMythologist:
I've looked through the commit and I think it's ready to be merged.

To avoid confusion, I used `.split(maxsplit=1)` when the `sep` argument is not passed into `.split` (i.e. it is splitting over any whitespace character).
In situations where the `sep` argument is already specified, I removed explicitly stating the `maxsplit` argument, and instead simply use `.split(".", 1)` as 1 will automatically be the `maxsplit` parameter.

kevoreilly:
Ready to be merged needs actual testing - there have been a lot of bugs recently due to lack of this, review or looking are not enough.

doomedraven:
ah nice, make sense @TheMythologist

TheMythologist:
I have applied the changes to my local CAPE instance and everything is running as per usual.

kevoreilly:
That is good - thanks for clarifying. If you could try and exercise the code that would be great.

I hope I don't appear ungrateful - I do appreciate all the code cleanup you have contributed recently.

TheMythologist:
Sorry but what do you mean by exercise the code?

Also no worries, I'm aware that creating such massive PRs is a headache for both of us.

kevoreilly:
I simply mean to execute the code that has changed with sufficiently variable input to test that it does not crash and works properly. I suggest this as it may not be sufficient to just 'run cape as usual' to get the code that has changed to be executed. Like fuzzing, good testing needs code coverage.

PR #682, opened 2021-12-29
Replace all `while 1` with `while True`

TheMythologist:
Minor fix to improvement readability (no code logic changed)


doomedraven:
thanks

PR #683, opened 2021-12-29
Convert `setdefault` and `append` to one-liner

TheMythologist:
Replaced the following:
```python
a.setdefault(x, [])
a[x].append(i)
```
with
```python
a.setdefault(x, []).append(i)
```

No code logic changed, all inputs and outputs remain the same.

### DIFF
--- a/modules/processing/parsers/CAPE/Loki.py
+++ b/modules/processing/parsers/CAPE/Loki.py
@@ -171,9 +171,6 @@ def config(filebuf):
 
     cfg = {}
     urls = decoder(filebuf)
-    if urls:
-        cfg.setdefault("address", [])
-
     cfg["address"] = [url.decode() for url in urls]
     return cfg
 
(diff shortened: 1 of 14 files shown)


modules/processing/parsers/CAPE/Loki.py
```diff
@@ -171,9 +171,6 @@ def config(filebuf):
 
     cfg = {}
     urls = decoder(filebuf)
-    if urls:
-        cfg.setdefault("address", [])
-
```
TheMythologist, line 176: These lines were removed because `cfg["address"]` will always be created in both situations:

Situation 1: `urls` is not an empty list -> `cfg["address"]` will be overwritten by `cfg["address"] = [url.decode() for url in urls]`
Situation 2: `urls` is an empty list -> `cfg["address"] = [url.decode() for url in urls]` will still create an empty list `[]` for `cfg["address"]`

doomedraven: make sense, thanks for comment

doomedraven:
ooo nice, i wasn't aware that it was working in that way, learned something new thanks

PR #685, opened 2021-12-29
Fix `.encode(hex)`

TheMythologist:
`.encode("hex")` no longer works in python3
It has been replaced with `.encode().hex()`

Unfortunately, I don't have any real way of testing this as I'm unsure of what the input and output types are (whether the function expects a byte-string or a string, and whether it returns a byte-string or a string), so I'll need some help on that.

Additionally, I am unsure of how to fix these for `.decode("hex")`, because they don't work in python3 as well.
An ugly fix that I currently have: replace `s.decode("hex")` with `bytes.fromhex(s.decode()).decode()`, which is very cumbersome.


doomedraven:
I saw that another day, and ypu know nobody complais means is deaf xode at least in static, as parsers for thosefamilies are dead

doomedraven:
Im afk rigth now but when i was porting i was using binascii.hexlify(data)

doomedraven:
@kevoreilly do you have any jrat sample to test or we can depricate it as nymaim?

doomedraven:
so fun part, about jrat, it not have a valid entry point `def config` so it won't called, also the name with prefix `_` won't match [yara](https://github.com/kevoreilly/community/blob/master/data/yara/CAPE/jRat.yar) + it doesn't work for any sample from https://bazaar.abuse.ch/browse/signature/jRAT/

so i will leave to decide to Kevin about those parsers

doomedraven:
there is also `.decode("hex")` in that jrat

doomedraven:
so i have found more parsers that weren't ported from py2, i have disabled them + disabled parsers that doesn't have proper entry point

PR #686, opened 2021-12-30
Fix child to_dict bug, optimize extrat path

N1neSun:
- Fix the to_dict bug that _sf_chlildren not work as expected.
- optimize if child is not available, the original file should be put into the list.


doomedraven:
hm interesting, i think this maybe could be sflock problem, could you share file to check that?

N1neSun:
f2d87361b01a4419d295f9b695527cf0aeb0fe40e43ab26445fa6c90a2b93ece, this hash can check it.

doomedraven:
thanks i was able to verify, interesting bug btw, never saw it before

PR #688, opened 2021-12-30
Add Elasticsearch reporting

CorraMatte:
Add class for Elasticsearch reporting support and small refactoring with common code for mongodb.
Edit the source code of the web app in order to display correctly the data.
Edit cleaners, admin, process, and sample_path utils.

Tested with a remote node with authentication and HTTPS.

### DIFF
--- a/conf/reporting.conf
+++ b/conf/reporting.conf
@@ -91,6 +91,10 @@ port = 9200
 # The report data is indexed in the form of {{index-yyyy.mm.dd}}
 # so the below index configuration option is actually an index 'prefix'.
 index = cuckoo
+# username =
+# password =
+# use_ssl =
+# verify_certs =
 
 [retention]
 enabled = no
--- a/dev_utils/__init__.py
+++ b/dev_utils/__init__.py
@@ -1,0 +1,3 @@
+# Copyright (C) 2010-2015 Cuckoo Foundation.
+# This file is part of Cuckoo Sandbox - http://www.cuckoosandbox.org
+# See the file 'docs/LICENSE' for copying permission.
--- a/dev_utils/elasticsearchdb.py
+++ b/dev_utils/elasticsearchdb.py
@@ -3,9 +3,11 @@
 # See the file 'docs/LICENSE' for copying permission.
 
 import datetime
-from elasticsearch import Elasticsearch
-from lib.cuckoo.common.config import Config
 import logging
+
+from elasticsearch import Elasticsearch
+
+from lib.cuckoo.common.config import Config
 
 repconf = Config("reporting")
 ANALYSIS_INDEX_PREFIX = f'{repconf.elasticsearchdb.index}-analysis-'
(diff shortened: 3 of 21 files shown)


kevoreilly:
Wow nice work.

doomedraven:
looks nice, thank you a lot! i have a question, with which ES version did you test? i think there we need few more `if`s as if no it will require elastic installed by default, but i will review once again in a bit and merge

doomedraven:
could you test again to ensure that my `if`s doesn't break please?

doomedraven:
ok im doing testing and i see latest version 7.16.2 works just fine thanks for that again

CorraMatte:
Sorry for the late response, I've tested with Elasticsearch stack version 7.15.2 and python library 7.16.1

doomedraven:
np, i have started a new PR to handle the deprication warnings, but im ES noob :) and having a lot of problems maybe some of them sound to you so you can help me, maybe we need to make a huge mapping idk but just in case leaving some parts that i have observed here

```
2022-01-03 21:09:29,407 [modules.reporting.elasticsearchdb] ERROR: RequestError(400, 'illegal_argument_exception', 'mapper [dropped.tlsh] cannot be changed from type [text] to [boolean]')

2022-01-03 21:39:42,798 [modules.reporting.elasticsearchdb] ERROR: RequestError(400, ‘illegal_argument_exception’, “can’t merge a non object mapping [signatures.data.regsvr32_exe] with an object mapping”)

elasticsearch.exceptions.RequestError: RequestError(400, 'illegal_argument_exception', 'Limit of total fields [1000] has been exceeded')
```

does it sound to you?

CorraMatte:
Yes, it's a problem related to the Elasticsearch mapping, at least the first 2. Elastic tries to guess dynamically the type of the field and it creates its mapping. When you try to insert a document with a different type for that field, it gives you that error.
The best solution is to create a mapping as I do when I create the index. I put some fields that gave me problems during the testing, but it would be better to put them all.

Also the last line is related to the mapping, but you need only to increase the field limit of the mapping.

doomedraven:
thank you
ok i see i guess is this one no?
so we can keep extending it
```
def get_analysis_index_mapping():
    return {
        "mappings": {
            "properties": {
                "info": {
                    "properties": {
                        "started": {"type": "date"},
                        "machine": {"properties": {"started_on": {"type": "date"}, "shutdown_on": {"type": "date"}}},
                    }
                },
                "network": {"properties": {"dead_hosts": {"type": "keyword"}}},
            }
        }
    }
```

CorraMatte:
Yes, you can export the current mapping dynamically created in order to do not create it from scratch. Then check if the fields are correctly mapped and solve the conflicts.

doomedraven:
thanks for tip, but looks like my problematic fields need to be added before ES trying to map them, the good part that is file info is the same under different keys

vinceplayer:
what version of elastic search is supported?

doomedraven:
7.16.2 works for me as you might see from my comments

PR #689, opened 2022-01-01
remove six library

doomedraven:
still have to fix few things related to strings and unicode


doomedraven:
@TheMythologist what do you think about the six removal changes?

TheMythologist:
Sorry for the late reply, but it looks good. There are still a couple of other files that use six:

Search for `from six`, e.g. `from six.moves import zip`.
The three files are:
- modules/processing/usage.py
- modules/processing/parsers/CAPE/REvil.py
- utils/tcpdumpwrapper.py

doomedraven:
Thanks, damn my grep didnt catch them, changing them

doomedraven:
ok i was grepping for `import six`, not `from six` thanks

PR #690, opened 2022-01-03
Use identity checking for `True`, `False` and `None`

TheMythologist:
Note the difference in changes for statements involving SQLAlchemy


doomedraven:
o thanks, learning more about sqlalchemy

PR #693, opened 2022-01-04
Add file name field for better web experience

N1neSun:
I don't know if this is a good pr. 

- Add a field to the database to display the real file name.
- need `cd /opt/CAPEv2/utils/db_migration && alembic upgrade head` and restart cape `systemctl restart cape`


doomedraven:
there is no such thing as good/bad PR, all of them are amazing, as all of us having different use cases

could you give me a bit more details please, so that would just show the original filename on web?

N1neSun:
Yes, this can show the original filename on web using the database filename field.

doomedraven:
i might be wrong so correct me if needed, what i see is basically that if we rename 
`path_to_extract` filename to `child.filename` we will get the same no?

N1neSun:
> 
Do you mean to use `child.filename` in `path_to_extract`?

doomedraven:
yes i think if we rename the file just after extract, instead of using path_to_extract, theorically that should do the trick to get the original filename

N1neSun:
yes,however, if there are characters such as unicode in the file name, there may be problems in subsequent analysis.

doomedraven:
hm, wouldn't that give problems when we store that in db? we can use `sanitize_filename` to ensure that it not gives problems later when renaming, but it can remove some chars, that probably will show a bit different name then?

N1neSun:
yes, there is no problems store in db.the purpose of this pr is to save the original file name in the database and display it on the web without being affected by sanitize_filename.

kevoreilly:
But if the original filename contains certain characters and is un-sanitized, will this not cause the entire web page to fail?

N1neSun:
> But if the original filename contains certain characters and is un-sanitized, will this not cause the entire web page to fail?

In fact, it should not be possible, because the django framework and python itself will help us clean up some abnormal strings, but I am not sure whether it can be bypassed.

N1neSun:
Considering that there may be some concerns about this function (actually static and pcap, etc. are all displayed origin filenames), I added enable in web.conf.

kevoreilly:
Too old to merge now - if you wish to revive this, please re-open.

PR #697, opened 2022-01-04
Allow config dirs to be provided that extend the default ones.

tbeadle:
Add a -c command-line option to cuckoo.py that can be provided multiple
times. This allows the *.conf files in this repo to be untouched by a
user. Instead the values that need to be overridden can be put in to one
or more other directories, whose path is passed via the -c option. The
values in those files will take precedence over the ones in the conf
dir.

In addition, for Config() objects, only read the files once, so that if,
say Config("processing") is created in multiple places, they'll both
refer to the same object and the conf files would only be read once.
Those files are also only read from the first time a section from that
config is referenced. This is because we so often create those Config
objects at module level scope and they could get imported before the
config_dirs have been set. So now, an initialize() classmethod needs to
be called on the Config class before any of those sections are
referenced.

A new AnalysisConfig() class was created to separate the distinction
between these types of "singleton" Config objects that can read from
multiple directories versus ones that just read from an analysis.conf
(or some other single file).


doomedraven:
nice, thank you, i will check that tomorrow

doomedraven:
it looks good, im just curious about that import in cuckoo.py, after that i will merge

tbeadle:
Actually, hang on. I may have found an issue. I'll be looking in to it this morning

tbeadle:
The moving of the imports in cuckoo.py is necessary since many modules create Config() objects during import (i.e. they're created as module-level variables), but we need to make sure `Config.initialize()` is called before they are created so that they know which directories should be processed. This is why there's the check in `Config.__init__` to verify that config_dirs is set...just in case somehow an import with this type of "global" got imported before `initialize` was called.

doomedraven:
so do you think is ready for merge? and now to access config from any submodule we need to init AnalysisConfig or which one?

tbeadle:
Yes, I believe it's ready for merge. The AnalysisConfig objects only need to be created when you're providing the path to a file that has ALL of the config inside of it--none of the extra directories provided on the command-line will be used to populate the object. In addition, the AnalysisConfig objects are not singletons. A new one will be created for each instance. This is because they are sample-specific, should be short-lived, and the same file name could be provided to many instantiations of it, but perhaps the cwd (the analysis directory) is different.

doomedraven:
thank you

PR #699, opened 2022-01-05
WIP: ES misses load data on demand as mongo does

doomedraven:


doomedraven:
@CorraMatte  if you don't care, could you help me/point me what would be the proper way to load only specific data from ES? it would be nice to be able load data on demand and not directly return the whole blob of data, that cuts load and timings

CorraMatte:
Hi, if I understood you are looking for the "_source" attribute in the Elasticsearch search() method. I use it in my pull request, try to have a look on it and let me know. [](https://elasticsearch-py.readthedocs.io/en/v7.16.3/api.html#elasticsearch.Elasticsearch.search)

doomedraven:
thank you

doomedraven:
oh i see that you implemented that already, what is missing as last bite is get counts of some fields like this, any tip im checking in docs but there many diff things and i don't have too much time right now, i would like to get count of array inside of `procdump`, example of mongo, any tip?

```
{"$project": {"_id": 0, "procdump_size": {"$size": {"$ifNull": ["$procdump.sha256", []]}}}},
```

doomedraven:
got it in a bit ugly way but it works :)

```
report[value] = len(es.search(index=get_analysis_index(), query=get_query_by_info_id(task_id),  _source=[f"{key}.sha256"])["hits"]["hits"][0]["_source"].get(key))
```

CorraMatte:
I wouldn't have a fastest solution, maybe I can have a look when I can to refactor that part of the code

PR #700, opened 2022-01-05
Allow environment variables to be referenced in config files.

tbeadle:
Allow config files to use values that are formatted like
`%(ENV:SOME_ENVIRONMENT_VARIABLE)s`. In certain deployment strategies,
this makes it easier to deal with secrets since they don't need to be
stored on disk anywhere.

The test case that has been added gives a normal use case.

If there are concerns over allowing this, perhaps it could be an opt-in feature via a command-line argument.


doomedraven:
a bit of docs would be good but im not even sure where to put that

PR #701, opened 2022-01-06
Update docs regarding the recent config file-related changes.

tbeadle:
Late documentation is better than no documentation!


doomedraven:
Thanks

PR #703, opened 2022-01-07
Revert recent config related changes

tbeadle:
The change that was recently made to allow for multiple config directories requires that Config.initialize() be called before any modules that have top-level Config objects created are imported. This was handled find in cuckoo.py, but there are plenty of other scripts that have not been addressed. I'll look at how to more elegantly address this problem.


doomedraven:
Thanks

PR #704, opened 2022-01-08
Convert to f-string for lib folder

TheMythologist:
Changes done:
1) Convert strings to f-strings (other than for logging)
2) Standardise logging (no full-stop at the end)
3) Removed redundant `print` statements (if logging is already done)
4) 2 changes in analyzer folder that were missed out

All files have been run through `black`

Note that files in `lib/cuckoo/pdftools` have not been modified as they are from someone else's repository, do let me know if you would like me to refactorise those code.


doomedraven:
thank you, about the 3rd part repos, if you have time you can do that and push to their repo. otherwise it will hard to keep it up to date and properly apply all your changes

PR #705, opened 2022-01-10
Fix url not work

N1neSun:
Because deleted `Task.tags.is_(None)` in fetch() of database.py,causes the url not be filtered into cuckoo's tasks.I add a tags for url.


doomedraven:
oh damn, nice catch

PR #707, opened 2022-01-12
Allow custom config dirs that extend the default ones.

tbeadle:
A new `custom/conf/` directory can be used to hold additional
configuration files that extend the ones in the top-level `conf`
directory. See the `custom/README.md` for more information.

In addition, for Config() objects, only read the files once, so that if,
say Config("processing") is created in multiple places, they'll both
refer to the same object and the conf files would only be read once.

A new AnalysisConfig() class was created to separate the distinction
between these types of "singleton" Config objects that can read from
multiple directories versus ones that just read from an analysis.conf
(or some other single file).

Allow environment variables to be referenced in config files.
Allow config files to use values that are formatted like
`%(ENV:SOME_ENVIRONMENT_VARIABLE)s`. In certain deployment strategies,
this makes it easier to deal with secrets since they don't need to be
stored on disk anywhere.


doomedraven:
thank you

doomedraven:
this doesn't work, it breaks analysis again, i will keep this on hold till i will have spare time to decently test it

PR #711, opened 2022-01-18
requires hard testing "Allow custom config dirs that extend the default ones.""

doomedraven:
Reverts kevoreilly/CAPEv2#710


doomedraven:
@tbeadle hey i just did a quick look and what is breaking is that you changed Config under analyzer folder but there is no such function/class is why it breaks the whole analysis

PR #716, opened 2022-01-19
json.dumps in python 3+ does not support encoding

LeanVel:
Full error
```
ERROR: Failed to run the reporting module "CompressResults":
  File "/opt/CAPEv2/utils/../lib/cuckoo/core/plugins.py", line 703, in process
    current.run(self.results)
  File "/opt/CAPEv2/utils/../modules/reporting/compressresults.py", line 38, in run
    cape_json = json.dumps(results[keyword], encoding='latin-1', ensure_ascii=False).encode('utf8')
  File "/usr/lib/python3.8/json/__init__.py", line 234, in dumps
    return cls(
TypeError: __init__() got an unexpected keyword argument 'encoding'
```

### DIFF
--- a/modules/reporting/compressresults.py
+++ b/modules/reporting/compressresults.py
@@ -37,7 +37,7 @@ class CompressResults(Report):
         for keyword in ("CAPE", "procdump"):
             if keyword in results:
                 try:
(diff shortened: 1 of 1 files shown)


doomedraven:
thanks, but bear in mind that compression disables search, as it won't be able to go and get what you want

PR #718, opened 2022-01-20
Revert Fix machine deadlock and optimization

N1neSun:
I realized that the last  fix #712 may not be good (may cause the task to be in the running state), I increased the check of the machine startup to ensure that some parameters and states of the machine startup are correct.


doomedraven:
thank you

PR #719, opened 2022-01-20
Revert 'Revert Fix machine deadlock and optimization'

N1neSun:
I'm terribly sorry that I may have messed up something (I didn't consider the multithreading), I think it might be better to leave it as-is. maybe restarting the service is the best solution.


doomedraven:
np :) thanks for trying to find the solution

PR #721, opened 2022-01-21
Update api tasks_view

N1neSun:
Update api `tasks_view()` to add detections, malscore and other fields from analysis information.


doomedraven:
thank you

PR #726, opened 2022-01-25
Add empty VTDL_PRIV_KEY to local_settings.py

Antyrael:
It took me a few seconds to figure out why virustotal uploading didn't work, after setting "[vtupload]" to "enabled = yes" in conf/web.conf.
This makes it easier to find after a fresh installation.


doomedraven:
thanks, i have fixed that to just replace that old depricated field with `VTDL_KEY` that is set in aux under `virustotaldl` i guess this requires a bit more simplification to bake them all together

PR #728, opened 2022-01-25
Mark certain strings as 'raw' strings.

tbeadle:
Stop the warnings about invalid escape sequences.


doomedraven:
i personally vote to remove those 2 first extractors, they are not even py3, and from 2016

PR #729, opened 2022-01-26
Add run statistics for task.

N1neSun:
add run statistics for task and add it to db.


doomedraven:
hello thank you, but i have suggestion in implementation

use reporting plugin under `reporting`, set order inside as mongodb.py +1 + reporting.conf conf configuration

also what is the goal? to show that on webgui? if yes i would suggest to move logic to mongo/es for counting it will be really faster then python

let me know if you think si better in current way

N1neSun:
hello, I just saw that there are related fields in postgreSql, but they are not implemented, so try to implement, as you said, it is indeed a good method to count these fields in the `reporting`, but if you write the database in the `reporting` ， Is it a good way to do it？

doomedraven:
at the end is depending of the usage of those data, to show on webgui i would prefer to have that in mongo/ES, but even to fill those in db, i think it still would be good to move code from plugins to reporting module as you can access db from there too. if you do that in reporting you not necessary need to write that to mongo/es, is just helps to get post processing data in final stage and you can do stats like this, you not even need to set order in reporting for this, i can show you example how it would look after my daily job

N1neSun:
ok, I will close this pr and try to implement in reporting(it would be better if you could give me an example, thanks a lot. : ) ).

doomedraven:
you don't need to close, you can just push changes to the same PR

for example, just see any reporting module, and you just need to import Database, see `modules/reporting/callback.py` how it inits db 
```
from lib.cuckoo.core.database import  Database
main_db = Database()
```

then you can can call your function, and one more thing could you add any parent key to this

```
instead directly write to root 
detail["dropped_files"]

do something like
detail["tasks_stats"]["dropped_files"]
```

doomedraven:
im fine with poc also, i always can change/improve what needed so don't worry about that

N1neSun:
oh, That's very nice, I understanded. thanks a lot again.

N1neSun:
I have changed it to reporting.

doomedraven:
thanks, i did small cleanup, but now it looks good as reporting module

N1neSun:
Also thank you, I learned a lot from this project and hope to help it get better.

doomedraven:
you are welcome, you did all the job, i just had to make small adjust, if you want to get the hidden gems, check `lib/cuckoo/common/abstracts.py`  classes of Processing, Signature, Reporting, it has common func that helps to write less code and do more ;)

N1neSun:
ok, i'll take a good look at it.

PR #732, opened 2022-01-31
Auxiliary and machinery modules cleanup

TheMythologist:
Some changes that I made:
- Removed redundant outdated Python checks
- Shifted code blocks to remove unnecessary indentations

Should be more readable in general, let me know if you disagree with any of the changes made

### DIFF
--- a/modules/auxiliary/sniffer.py
+++ b/modules/auxiliary/sniffer.py
@@ -151,13 +151,9 @@ class Sniffer(Auxiliary):
 
             subprocess.check_output(
                 ["scp", "-q", f"/tmp/{self.task.id}.sh", remote_host + f":/tmp/{self.task.id}.sh"],
-                stdout=subprocess.DEVNULL,
-                stderr=subprocess.STDOUT,
             )
             subprocess.check_output(
                 ["ssh", remote_host, "nohup", "/bin/bash", f"/tmp/{self.task.id}.sh", ">", "/tmp/log", "2>", "/tmp/err"],
-                stdout=subprocess.DEVNULL,
-                stderr=subprocess.STDOUT,
             )
 
             self.pid = subprocess.check_output(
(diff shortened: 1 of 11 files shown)


doomedraven:
looks good, thank you, just few things

1. `return None` is the same that just `return` as far as i know
2. about `subproces.check_output` if we remove return value storage, it will print to stdout, so i have 2 suggestions.
    * add `stdout=DEVNUL` <- i prefer this one
    * or `_ = subprocess.check_output`
    
what do you think?

TheMythologist:
I think we should stick to using `return None` because the function does return None instead of simply returning (which I would use only in places to break out of loops). I will add subprocess to redirect its stdout to `DEVNULL` , thanks for the input :)

doomedraven:
ok make sense, thank you

TheMythologist:
Actually, before merging, I took a look and I don't think `subprocess.check_output` prints to stdout, I'll make changes accordingly

doomedraven:
well i tested before make comment, and it gave me outout in stdout

TheMythologist:
May I know what's the output/command you're running? I can't seem to replicate on my end

doomedraven:
to simulate ssh and cat

```
>>> subprocess.check_output(["ssh", "<removed>",  "cat", "/etc/passwd"])
b"root:x:0:0:root:/root:/bin/ba<removed>
```

doomedraven:
anything that is not stored to output will print, we could change it from check_output as it forces subprocess to return output and not just return comamnd, but with stdout=DEVNUL if should be fine

TheMythologist:
Ahh I see it's because you're running it in the Python interpreter, which will print the return value of the function you're running. Try running the command `subprocess.check_output(["ssh", "<removed>",  "cat", "/etc/passwd"])` in a python file and running the python file itself from the command line and you won't see the output!

doomedraven:
ok yes you are right

doomedraven:
i have cleaned a bit more vbox machinery, but i hope nobody uses that as we not even care/supports that https://docs.oracle.com/en/virtualization/virtualbox/6.0/user/vboxmanage-debugvm.html

PR #736, opened 2022-02-02
Changes to `pyproject.toml`

TheMythologist:
Edit `pyproject.toml` to include .py_disabled and fix compatibility issues between `black` and `isort`
As a result, we no longer have to skip file "lib/cuckoo/common/integrations/capa.py" for `isort`

Also ran black (and isort) and all the files again


TheMythologist:
By default, isort will look at `pyproject.toml` for configuration settings, so we don't need to specify the settings in the github workflow

doomedraven:
thank you

PR #737, opened 2022-02-02
Remove redundant python version checking

TheMythologist:
_Note_: Feel free to revert d1544bbba5bef11be0c7a419de01bfec0559fa67 if you feel that we should not edit 3rd party code


doomedraven:
thats fine thanks, i just hope the authors of those projects would like the code update/cleanup as i love

PR #738, opened 2022-02-02
Run latest black version and update github workflow

TheMythologist:
_Note_: No need to independently install each linter (it's a part of each github workflow)


doomedraven:
thank you a lot for work

PR #741, opened 2022-02-02
Cleanup of modules/processing/parsers/CAPE folder

TheMythologist:
Some changes that I made:
- Removed redundant outdated Python checks
- Shifted code blocks to remove unnecessary indentations
- Mainly minor/shifting of code for it to be more readable (hopefully) (no code logic was changed)

I apologise if this commit is relatively large, but I have no idea how to split it up even further

### DIFF
--- a/modules/processing/parsers/CAPE/Azorult.py
+++ b/modules/processing/parsers/CAPE/Azorult.py
@@ -17,6 +17,9 @@ import struct
 import pefile
 import yara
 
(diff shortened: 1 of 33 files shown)


doomedraven:
wow, thats fine if that is big but when change is you know more less similar, is easier to review, than change tons of different things, but my main point is to core code, that is hard part to review, the rest is kinda ok

PR #752, opened 2022-02-08
Update Remcos config parser

enzok:


doomedraven:
i have moved from mwcp to pure cape, tested it with my extractor, only 1 sample was handle by this one

i see that many fields contains binary data as `'Unknown56': b'0\x81\xfe0\x81\xa6\xa0\x03\x02\x`

PR #763, opened 2022-02-14
static 2 integrations

doomedraven:


doomedraven:
the idea of this PR is to be able to reuse parsers, as for example if initial binary is lnk file, but it downloads another file, we won't parse that, with this change we can parse any other file of the same format

doomedraven:
Screenshot 2022-02-16 at 17 59 20

PR #770, opened 2022-02-17
mongo_update to mongo_update_one

enzok:
mongo update() is deprecated and seems to crash with new pymongo


doomedraven:
thank you, what version of pymongo ? to put in requirements

PR #773, opened 2022-02-20
Add sha256 hash to procmemory result dict

enzok:


doomedraven:
i have updated it as we already have instance of File, but now i undersand that is was missed in views, as i checked for file but not for get_all, so thanks

PR #792, opened 2022-03-06
fix access of sample without assignment

antihax:
URL samples fail to process as sample.sha256 is not assigned. 
There should be nothing to copy.


antihax:
Resolves the error:

```
2022-03-06 22:11:10,797 [modules.reporting.mongodb] DEBUG: Deleted previous MongoDB data for Task 8
2022-03-06 22:11:10,885 [root] INFO: Task #8: reports generation completed
  File "utils/process.py", line 245, in autoprocess
    copy_origin_path = os.path.join(CUCKOO_ROOT, "storage", "binaries", sample.sha256)
UnboundLocalError: local variable 'sample' referenced before assignment
```

doomedraven:
thank you

PR #793, opened 2022-03-07
Fix logic for finding if a detection was already seen.

tbeadle:
The line `if not detection in block["details"]:` would never evaluate to
True, because block["details"] contains references to dicts created in
previous invocations of this function--that statement does not actually
compare the contents of the dicts themselves, but only that they're the
same object.

Also, bail out of the loop when we found the family and use python's
for/else construct instead of tracking `found` ourselves.

QUESTION: I was curious about the decision to use lists for these objects (results['detections'] and the 'details' instead of just using dicts that key off the family and detected_by, respectively. I'm guessing there was a reason--it just seems like it would've been simpler than having to iterate over these lists to find items.

### DIFF
--- a/lib/cuckoo/common/utils.py
+++ b/lib/cuckoo/common/utils.py
@@ -646,14 +646,13 @@ def store_temp_file(filedata, filename, path=None):
 
 def add_family_detection(results: dict, family: str, detected_by: str, detected_on: str):
     results.setdefault("detections", [])
-    found = False
     detection = {detected_by: detected_on}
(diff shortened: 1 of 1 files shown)


doomedraven:
about the question i already answered in general in slack, the problem is search become hackish, that was how i initially designed that. but the main problem that there can be 2 files with the same match or more than 2, and there was more and more complications, if you have better idea how to implement this, im more than happy to improve this

thanks for improvements

PR #796, opened 2022-03-09
Code cleanup: Part 1

TheMythologist:
(Starting from the number 1 because I didn't keep track starting from the previous code cleanup commits)

Note: I'll start to include type-hinting in the cleanups from the next PR onwards

Let me know if there's anything wrong/erroneous, or anything that you disagree with


doomedraven:
thank you, i was missing you :)

PR #797, opened 2022-03-10
Cleanup of admin, agent and dev_utils folder

TheMythologist:
Let me know if there's anything you disagree with

* Introduce type-hinting on a larger scale

### DIFF
--- a/admin/admin.py
+++ b/admin/admin.py
@@ -25,6 +25,7 @@ import sys
 from hashlib import sha256
 from queue import Queue
 from threading import Thread
+from typing import Iterable
 from urllib import urlparse
 
 import urllib3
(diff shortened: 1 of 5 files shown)


doomedraven:
looks good thanks, just one thing, this isn't coming by default with python no `from pydantic import ipaddress` ? at least i don't have it on my 3.8.10, but `ipaddress ` does come, maybe is better to switch to `ipaddress` ? what do you think?

doomedraven:
there is another problems, as we convert methods from list to tuple, we need to add `,` at the end, otherwise it loop over strings, 

``` method is GET os is looping over GET string
  File ".\test.py", line 218, in <module>
    @app.route("/")
  File ".\test.py", line 108, in register
    self.routes[method].append((re.compile(f"{path}$"), fn))
KeyError: 'G'
```

PR #806, opened 2022-03-18
Whitelist 8.8.8.8 and 1.1.1.1 as hosts.

tbeadle:
Also allow comments in the domain and IP whitelist files to be on the
same line as the entry (e.g. `1.1.1.1  # such and such`) and don't read
the whole file in to memory at once.


doomedraven:
thanks

PR #814, opened 2022-03-29
swap django.conf.urls.url for re_path

nbargnesi:
The url function was deprecated in Django 3.1 and left in as in alias to `re_path`.


doomedraven:
oh nice, thank you

doomedraven:
there is django 4 as far as i know, one day would be nice to check what is new, but as far as i remember django 3.2 is LTS

PR #834, opened 2022-04-26
Bug fix for

enzok:
Reject forwarding to a certain network segment


doomedraven:
thanks

qux-bbb:
I know reject_segments should be initialized, but `if routing.routing.reject_segments == "none":` cannot be deleted.  
This is proper:  
```python
if routing.routing.reject_segments != "none":
    self.reject_segments = routing.routing.reject_segments
```

PR #848, opened 2022-05-10
Add AWS support

JaminB:
- added `aws.py` adapted from [here](https://github.com/CheckPointSW/Cuckoo-AWS).
- added corresponding `aws.conf`
- updated requirements*


JaminB:
Split into two PRs:
- https://github.com/kevoreilly/CAPEv2/pull/848 (AWS Machinery)
- https://github.com/kevoreilly/CAPEv2/pull/849 (Reversing Labs Processor)

doomedraven:
thank you

kevoreilly:
Thanks a lot.

Out of interest, how well does it work on AWS - does it scale well, any vm detection issues, etc?

JaminB:
I've done limited testing (only scaled up to 5 instances concurrently), but so far so good. I'll report back if I run into any issues as I scale up.

I haven't run into it personally, but if the `cuckoo` crashes during analysis you could be left in a situation where you have several orphaned EC2 instances running (as this module manages its own autoscaling internally) so suggest keeping the `dynamic_machines_limit` low or baby-sitting your EC2 environment until you feel confident your configuration is solid.

Only other kind of tricky part is the routing to get traffic back to the CAPE instance for packet-capture. All of that has to be done with routing tables inside your VPC. I eventually accomplished it by creating an AMI image always deploys inside the the same subnet, and then building an explicit routing table for that subnet that routes traffic to CAPE. After, that all of the per-analysis routing should be the same (although I have yet to implement it).

Will be sure to push any fixes if I find any bugs.

PR #849, opened 2022-05-10
Add reversinglabs processor

JaminB:
- adds support for reversinglabs API (note tested on A1000 appliance)
- Added summary tab to display results in web-ui


doomedraven:
thank you

PR #855, opened 2022-05-23
"CAPE Yara" format display

qux-bbb:
All "CAPE Yara" use same display style.  
Not "meta.name" rather than "name".


doomedraven:
thank you

PR #863, opened 2022-05-27
Fix condition for adding tasks for extracted files.

tbeadle:
Note that I'm not totally positive that this should be `only_extraction`, but I am pretty sure it's not supposed to be `static_extraction` because, since that's a function, it will always be truthy.


doomedraven:
That is required for checkbox try extractwithout vm, if extrqcts, do not submit to vmas far as i remember, thanks i had to look into that after pto

PR #864, opened 2022-05-27
Refactor some code for handling cape_type/cape_name and change how static_config_parsers is called

tbeadle:
- Pass the cape_name to static_config_parsers instead of the rule name.
  The cape_name is derived from the cape_type meta field in the matching
  rule. It is the part of the name before "Payload," "Loader," or
  "Config."
- Use this "cape_name" as the detection family instead of using the rule
  name.

Rules currently in data/yara/CAPE/ have a cape_type that
follows that convention so that there *should* be no change in detections.
One issue is:
 - EternalRomance: This has a cape_type of "EternalRomance Exploit". I'm
   not sure how this should be handled since it does not end in
   "Config", "Payload", or "Loader".


doomedraven:
Thanks Tommy, yes exploit is diff, and we need to add some custom keyword that allows to user add their own detection for visualization but well is a good start so far

PR #866, opened 2022-05-31
Update QakBot.py

yevhenprotsenko:
QakBot extractor is generating some 'keys' which are integers. This makes the JSON library fail and it also prevents the proper insertion into MongoDB.


kevoreilly:
Thank you!

PR #875, opened 2022-06-07
Use Poetry

TheMythologist:
# Use [Poetry](https://python-poetry.org/) to handle Python dependencies

I hope this will be merged (even though it can be tedious/troublesome to get all staging/production environments to use this) because the current dependency version control that is in place now is almost non-existent.

## Advantages:
- Significantly easier to install Python dependencies
```bash
sudo -u cape pip3 install poetry
sudo -u cape git clone https://github.com/kevoreilly/CAPEv2.git /opt/CAPEv2
cd /opt/CAPEv2
# Change the home directory of cape to /opt/CAPEv2
sudo usermod -d /opt/CAPEv2 cape
sudo -u cape poetry install
# The poetry environment will be stored in /opt/CAPEv2/.cache/pypoetry
```
- Significantly easier to handle Python dependency trees
- Significantly easier to update Python dependencies
  - Simply run `poetry update <package_name>`

## Disadvantages:
- Requires you to know poetry
- Potentially automated steps to install CAPE will be broken (e.g. https://github.com/doomedraven/Tools)

## Useful Poetry knowledge (skip if you already know how to use Poetry):
- **NEVER** directly edit the `poetry.lock` file unless you really know what you're doing
- To add dependency: `poetry add <package_name>`
  - Specific package versions and git repositories are supported as well
  - Reference: https://python-poetry.org/docs/cli/#add
- To update a dependency: `poetry update <package_name>`
  - You can update all dependencies as well via `poetry update` (but note that this may break with massive updates)
  - Reference: https://python-poetry.org/docs/cli/#update


TheMythologist:
Do let me know if you have any concerns / issues with this PR and I'll try to address them ASAP

themanifold:
Here is my concern about this:

Poetry, though great, is not as popular as plain old pip, so if you lock people into using that package manager if they want to use cuckoo there will be some annoyed users.

Proposed solution:

I am not familiar with poetry, but is there some way in which you could also support pip too? Maybe that is already a thing but I see a `poetry.lock` and no .txt files. Could you maybe add a step where you use that lock file to generate a .txt file and store that in git too?

I think I can see a poetry option here:

```
poetry export --format, requirements.txt --output, requirements.txt
```

TheMythologist:
> Proposed solution:
> 
> I am not familiar with poetry, but is there some way in which you could also support pip too? Maybe that is already a thing but I see a `poetry.lock` and no .txt files. Could you maybe add a step where you use that lock file to generate a .txt file and store that in git too?
> 
> I think I can see a poetry option here:
> 
> ```
> poetry export --format requirements.txt --output requirements.txt
> ```

Yup sure that sounds like a good idea so we leave it up to the user to decide whichever dependency management tool to use, I'll add it to the PR

doomedraven:
amazing, thank you both, can maybe you reinclude the requirements.txt in PR and i will merge

doomedraven:
i have tested `poetry export --format requirements.txt --output requirements.txt` works just fine. merging

PR #879, opened 2022-06-08
refactor: remove __future__ imports

TheMythologist:
The `__future__` module is only relevant in python 2 and does not achieve anything in python 3 projects


doomedraven:
thank you

PR #890, opened 2022-06-09
Implement new features in utils/community.py

TheMythologist:
New features:
- Support downloading of community modules from gitlab (useful for forks located on gitlab)
- Support downloading of private repositories via the usage of private access tokens (useful for private forks)

### DIFF
--- a/utils/community.py
+++ b/utils/community.py
@@ -85,15 +85,21 @@ def mitre():
     mitre.update()
 
 
-def install(enabled, force, rewrite, filepath):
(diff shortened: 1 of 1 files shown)


doomedraven:
Thank you, about private i had in my ToDo, but no time :)

doomedraven:
i have to add the blacklist for community, as some yaras, sigs, etc from community i don't want on my side. i will try to add it today, but no promises as i will be afk for rest of the week

PR #896, opened 2022-06-10
Fixing bug where boolean values were received as strings

cccs-kevin:
As per the QueryDict [documentation](https://docs.djangoproject.com/en/4.0/ref/request-response/#django.http.QueryDict), values received from the POST body are interpreted as strings.

For example, if I submit a file to the REST API with a parameter that contains boolean value, such as `enforce_timeout`, set to `False`, this will be received as `"False"` by the REST API here 

https://github.com/kevoreilly/CAPEv2/blob/768334e13515f6700a49c3422d6279991175041c/web/apiv2/views.py#L256

Then this value is converted to a boolean here 

https://github.com/kevoreilly/CAPEv2/blob/768334e13515f6700a49c3422d6279991175041c/lib/cuckoo/common/web_utils.py#L1036

This will always evaluate non-empty strings to True, even if you set the value to be False in the submitter client.

See my proposed fix to address this bug :)

### DIFF
--- a/lib/cuckoo/common/web_utils.py
+++ b/lib/cuckoo/common/web_utils.py
@@ -1015,6 +1015,22 @@ def force_int(value):
         return value
 
 
+def force_bool(value):
+    if type(value) == bool:
+        return value
+
+    if not value:
+        return False
+
+    if value in ["False", "false", "FALSE"]:
+        return False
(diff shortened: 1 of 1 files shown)


doomedraven:
thank you

PR #907, opened 2022-06-13
feat: priority in auxiliary modules

TheMythologist:
This is already present under linux packages, not sure why it wasn't brought into the windows ones.


doomedraven:
One small thing, maybe we could use getattr or something like that as if someone has custom module it will break module till they update, and being inside VM it will take them a while to update it. so maybe put 0 as default? what do you think? but apart of that looks good to me

doomedraven:
ah nevermind i see that abstracts has 0, sorry just wakedup xD

PR #917, opened 2022-06-14
Allow extracted files that match yara rules with a cape_type to trigger config extraction.

tbeadle:
*Note*: I am not familiar with how the extracted_files attribute of file_info is used. All I know is that I downloaded a SocGholish sample from a CAPE sandbox and resubmitted it. When it got to this point in the code, the cape_yara attribute was empty, but the file that matched the yara rule was in extracted_files and that had a yara rule match. Without this code change, no config extraction would be done.

tl;dr - Please make sure that this change makes sense. I only have a limited amount of test data.

Is the 'data' attribute of the extracted_file always a str (as it was in my SocGholish example) or could it be a 'bytes' object?

### DIFF
--- a/modules/processing/CAPE.py
+++ b/modules/processing/CAPE.py
@@ -249,17 +249,11 @@ class CAPE(Processing):
         # Process CAPE Yara hits
 
         executed_config_parsers = set()
-        for tmp_data, hit in (
-            [
-                (file_data, yara)
-                for yara in file_info["cape_yara"]
-            ]
-            + [
-                (extracted_file["data"].encode(), yara)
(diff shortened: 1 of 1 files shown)


doomedraven:
good one, im personally use yara_match more than this but is a good to have. about data, i guess that is depending of the proper tool that extracts it so we would need to ensure that the data always is the same

tbeadle:
The thing I submitted was a zip file (password "infected") containing the file at https://www.virustotal.com/gui/file/4f58e91704efb836e464392c5f6afaeb232aac5c8e665ba28cdf7b4b626ace0a

doomedraven:
thank you

PR #929, opened 2022-06-17
Convert task id to int, change disable message, enforcing line length format

cccs-kevin:
- Task ID is parsed from the REST API as a string, but must be an integer for the Mongo connection to delete it https://github.com/kevoreilly/CAPEv2/blob/d426faabb6de16d4692cab651673d074f7c69bb5/dev_utils/mongodb.py#L141
- Looks like a copy-paste error in the "disabled" message for apiconf.taskreport endpoint.
- Enforcing line length of 120 characters


doomedraven:
line size is 132 ;)

PR #930, opened 2022-06-17
Updating the line length

cccs-kevin:
REL: https://github.com/kevoreilly/CAPEv2/pull/929#issuecomment-1158987959


doomedraven:
now you know better the docs than I :)

PR #936, opened 2022-06-20
Fixing REST responses, adding helpful log, fixing path check

cccs-kevin:
- `.format` wasn't being used
- fixed incorrectly named `esp` variable
- added warning log if there is no `lite.report` and `lite` is requested
- merging conditional checks


doomedraven:
thank you

PR #939, opened 2022-06-20
Add ability to run multiple dll functions by name, either by specification, or by discovery

cccs-kevin:
Modifying the `dll.py` package to allow the ability to run multiple DLL functions if specified by name in the `function` option, or if `pefile` is installed on the guest then we can dynamically find libraries to run.

### DIFF
--- a/analyzer/windows/modules/packages/dll.py
+++ b/analyzer/windows/modules/packages/dll.py
@@ -3,11 +3,14 @@
 # See the file 'docs/LICENSE' for copying permission.
 
 import contextlib
+import logging
 import os
 import shutil
 
 from lib.common.abstracts import Package
 from lib.common.common import check_file_extension
+
+log = logging.getLogger(__name__)
 
 
 class Dll(Package):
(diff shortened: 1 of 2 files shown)


doomedraven:
Hello, that won't work for many cases, i did that in past, and i have code if you wants that extracts exports with pefile on server side and pass them as parameter and we have cases where dll performs unpacking and exposing exports after dllmain. + we saw huge performance hurting with DLL detonation

cccs-kevin:
> Hello, that won't work for many cases, i did that in past, and i have code if you wants that extracts exports with pefile on server side and pass them as parameter and we have cases where dll performs unpacking and exposing exports after dllmain. + we saw huge performance hurting with DLL detonation

Hey @doomedraven, that makes sense. I'll close the ticket

doomedraven:
here is how i had added that

https://github.com/kevoreilly/CAPEv2/blob/master/lib/cuckoo/core/scheduler.py#L238
```
options["exports"] = PortableExecutable(self.task.target).get_dll_exports()
```

so you just need to add something like this. as you already have list of exports so you don't need to add pefile inside of the vm
```
dll_exports_num = min(len(self.config.exports.split(",")), 8)
if dll_exports_num:
            function = self.options.get("function") or f"#1-{dll_exports_num}"
```

cccs-kevin:
I'm briefly reopening this ticket to update the PR with a commit before I close it again.

This PR is for those who WANT the default behaviour of the `dll.py` package to execute all available exports (up to a limit), and for those who WANT to be able to request function names to be used instead of ordinal numbers. 

Obviously this is not for everyone, as @doomedraven has mentioned.

With their suggestions, `pefile` is no longer required on the guest and the exports are discovered server-side.

cccs-kevin:
> this is fine to merge now no?

Yes I believe so

doomedraven:
thank you

PR #942, opened 2022-06-21
Changing resultserver file upload logs clearer

cccs-kevin:
If a file was uploaded, we didn't know what file it was.


doomedraven:
We had some discussion in one of the issues to not use f-strings for logging due to performance, im on phone, but if you interested to see that discussion it shouls be easy to find, but yes this is a good pr, but can ypu change it to use old format instead of f-steings and i will merge

PR #944, opened 2022-06-21
Adding filepickup aux module

cccs-kevin:
This module allows you to pick up a file that a sample may have dropped, even with `free=yes`.


doomedraven:
interesting feature, thanks

PR #951, opened 2022-06-22
Temporary fix for monitor not uploading behaviour logs

cccs-kevin:
I'm not sure why this is happening, but if we don't start every auxiliary module then the CAPE monitor does not upload behaviour logs to the resultserver. This is a temporary fix until someone is able to find out why 🤯 

Possibly related to the `AUX_ENABLED` variable?


doomedraven:
Could be, im not sure neither, my brain is off today already

PR #952, opened 2022-06-22
helper script for installing libvirt-python via poetry run

VVelox:
So `poetry run` can't take complex commands, but it can take a path to a helper script that will install the required libvirt-python depend. :)

### DIFF
--- a/extra/poetry_libvirt_installer.sh
+++ b/extra/poetry_libvirt_installer.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# run this via...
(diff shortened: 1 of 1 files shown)


VVelox:
Based off of `https://github.com/kevoreilly/CAPEv2/discussions/949` with the libvirt-python version chosen the same as `https://github.com/doomedraven/Tools/blob/master/Virtualization/kvm-qemu.sh`.

VVelox:
> thank you for this, today too much load to handle everything and my brain is off

Null sheen! Happy to help and I wish you the best with that, dude. :)

PR #954, opened 2022-06-22
Adding support for inetsim port redirection by range

cccs-kevin:
Adds support for port redirection by range


doomedraven:
oh i remember i had that for linux cuckoo

PR #960, opened 2022-06-27
Initialize sorted key

enzok:


doomedraven:
amazing, thanks for fix

PR #961, opened 2022-06-27
Removing unused folders, reusing RESULT_UPLOADABLE

cccs-kevin:
All folders listed under `RESULT_UPLOADABLE`:
https://github.com/kevoreilly/CAPEv2/blob/master/lib/cuckoo/core/resultserver.py#L46:L61

other than `report` should be created in this method here:
https://github.com/kevoreilly/CAPEv2/blob/master/lib/cuckoo/core/resultserver.py#L336:L350
?

Also it looks like `buffer` and `extracted` are unused in CAPE.


doomedraven:
buffer is old stuff that comes from cuckoo-modified that was something to do with exploits

PR #966, opened 2022-07-01
Include the ID of the task being processed in all log messages.

tbeadle:
When --parallel is used by process.py, it can be hard to tell what task
ID is associated with many of the log messages. It also puts a burden on
other parts of the code to include the task ID in log messages to make
them useful.

Instead, since only 1 task is being handled in a particular process from
the pool at a time, just update the formatter used for the log messages
to include the ID of the task.


tbeadle:
I'm not sure how to address the issue with not finding black in the CI pipeline.

doomedraven:
Oh thats cool i had this in todo, thabks

PR #967, opened 2022-07-01
Fix bug in parsing ranges of ID's in process.py.

tbeadle:
If a valid range of ID's was given on the command line, parse_id
would raise an exception. This fixes that.


tbeadle:
I'm not sure how to address the issue with not finding black in the CI pipeline.

doomedraven:
Just run "isort ."

PR #972, opened 2022-07-05
feat: display PE overlay information

TheMythologist:


doomedraven:
did you test this `data-url="/analysis/load_files/{{analysis.info.id}}/static/pe/overlay"` i think that will be problematic part as we need to edit `urls.py` and views where we handle ajax request as:
1. urls expect only 1 keyword after task id that is set as. category
2. in ajax handler in `views.py` we need to use that category. so maybe change `static/pe/overlay` to `pe_overview` ?

TheMythologist:
The url returns 404 but the overlay data tab still displays the information accurately

doomedraven:
ok, well a bit later today i will do mods to not fetch ajax part and go directly to that data

TheMythologist:
Never mind I got it :)

PR #973, opened 2022-07-05
feat: floss

TheMythologist:


doomedraven:
hello just to leave here few notes to consider. In future we must to move this to python implementation, maybe later they will post more examples, to reduce HDD I/O as in past that already was bottleneck for me on gran scale processing. 

about `subprocess.run` im not familiar with it so much, but recently i saw that due to missed `communicate()` it was waiting in `sniffer.py` for someone to fetch the output in some rare cases and that was blocking the rest. With processing this shouldn't happens but just leaving this as note in case someone see rare stucks

doomedraven:
ah btw config option missed. Can you add it please?

PR #976, opened 2022-07-05
Adding periodic logger to enhance debugging of pending tasks, locked machines, etc

cccs-kevin:
This may not be for everyone since it adds 4 additional DB calls every ten seconds, but it is crucial to our operations to be able to determine the backlog in CAPE and the status of locked vs unlocked machines.


doomedraven:
Can it be conf on/off? And then i will merge it directly

cccs-kevin:
> Can it be conf on/off? And then i will merge it directly

Added!

PR #983, opened 2022-07-06
Miscellaneous web changes

TheMythologist:


doomedraven:
hey, thanks for all your PR. i have question about this one, about bingraph, does it removes bingraph on demand button?

TheMythologist:
Nope it doesn't, it just fixes the font issue

doomedraven:
i will need to double check as i see `elif` removed

PR #994, opened 2022-07-07
Updating the processing of sysmon data

cccs-kevin:
Updated based on https://github.com/kevoreilly/community/pull/291

Side note.... should this processing module be moved to the community repo?


kevoreilly:
It might be better to put this in community repo - this was created to remove the burden of potential breakages to the working core from contributions which we can't directly support... The downside is there might be some who miss out if they don't install/update community... Ultimately if something is seen as 'core'/stable enough then it could be moved to the main repo too...

cccs-kevin:
@doomedraven I think you still have to merge this so that the sysmon processing module is removed from the core

doomedraven:
good point, thank you

PR #998, opened 2022-07-11
Fix bug in CAPE processing.

tbeadle:
Some samples may match multiple yara rules that have a cape_type
defined. If the first one has a config extracted but the last one
doesn't, then, previously, self.cape["configs"] would end up having a
an entry with the top-level key of the latter cape_type and its value
would be a dict whose key is the first cape_type and its value would be
the extracted config. This fixes that problem by making sure to add each
config to self.cape["configs"] when it is found instead of waiting until
the very end. This is because the 'cape_type' variable can change as it
goes through iterations of the loop.

I factored the code a little bit to try to make it more readable.

This also fixes a problem where not all cape_names would get added as
family detections. Again, this was because we were only using the last
cape_name that was found.

Testing: I detonated a sample that previously had a config that ended up being `[{"Metasploit": {"CobaltStrikeBeacon": {...cobalt strike beacon config...}}}]` and the only detection was "Metasploit". Now it stores the config as `[{"CobaltStrikeBeacon": {...cobalt strike beacon config...}}]` and has "CobaltStrikeBeacon" AND "Metasploit" as detections.
I don't have a PlugX sample to confirm the change there.

### DIFF
--- a/modules/processing/CAPE.py
+++ b/modules/processing/CAPE.py
@@ -94,10 +94,27 @@ multi_block_config = ("SquirrelWaffle",)
 class CAPE(Processing):
     """CAPE output file processing."""
 
+    def add_family_detections(self, file_info, cape_names):
+        for cape_name in cape_names:
+            if cape_name != "UPX" and cape_name:
+                if processing_conf.detections.yara:
+                    add_family_detection(self.results, cape_name, "Yara", file_info["sha256"])
+            if file_info.get("pid"):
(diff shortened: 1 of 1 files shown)


tbeadle:
I'm not sure there are ModuleNotFound errors while running tests.

doomedraven:
don't worry about those tests. thank you

kevoreilly:
Thank you!

PR #1002, opened 2022-07-12
Fix KeyError in virustotal integration.

tbeadle:
We were checking virustotal to see if it has the 'detection' key, but
then attempt to use vt_response, which might not have that key. Also,
the 'sha256' key does not exist in vt_response, but it does in
virustotal.


doomedraven:
thanks

PR #1009, opened 2022-07-14
Generalizing log sub-format, fixing cuckoo.log per-task

cccs-kevin:
The `init_logger` method seems to be half-migrated from Cuckoo. The PR implements it, and populates the cuckoo.log per-task.

Also tried to make the sub-format for each log in the Guest and Resultserver the same:
`Task #<task-id>: blah blah (id=<vm-id>, ip=<vm-ip>)`

### DIFF
--- a/cuckoo.py
+++ b/cuckoo.py
@@ -66,12 +66,14 @@ def cuckoo_init(quiet=False, debug=False, artwork=False, test=False):
         except KeyboardInterrupt:
             return
 
-    init_logging()
-
     if quiet:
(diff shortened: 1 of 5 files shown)


doomedraven:
thank you, one question, did you test this change? asking before go to testing

cccs-kevin:
> thank you, one question, did you test this change? asking before go to testing

Yes it is running in my production environment

doomedraven:
thanks for confirmation

PR #1012, opened 2022-07-16
make poetry play nice with utils/submit.py

VVelox:
So was just sitting down for writing a perl module for fetching and shoving a lot of drek into CAPE and I noticed that utils/submit.py was borked post move to poetry.

Updated the interpreter for the script to `#!/usr/bin/env -S poetry run python3`, which will automatically invoke poetry to run the script.

### DIFF
--- a/utils/submit.py
+++ b/utils/submit.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env -S poetry run python3
 # Copyright (C) 2010-2015 Cuckoo Foundation.
 # This file is part of Cuckoo Sandbox - http://www.cuckoosandbox.org
 # See the file 'docs/LICENSE' for copying permission.


doomedraven:
Hello, not everyoen uses poetry, is why we generate requirements.txt. so changing this will impact on everyone who is not on poetry, i still prefer them to run it as `poetry run python3`

TheMythologist:
Maybe we can standardise by removing all the shebangs?

VVelox:
@doomedraven Submitted this as I had it break after a recent update and could not get it to work with out it.

A better solution, should any one have it working outside of that, may be a poetry wrapper for it then. A second script sitting there that can be invoked to automatically call the other via poetry.

VVelox:
@TheMythologist Personally I would say that would just be ultra annoying.

doomedraven:
annoying why? sorry guys didn't have time recently for this. this week is still pretty busy, i guess removing the shebangs should work, but in general if you with poetry you just need to run with `poetry run python submit.py` and if you don't use poetry you just use it as `python3 submit.py`. so i don't know what is the issue there, i will need to test it more in deep, but i would say nobody of us using submit.py as we do not work on server side to add analysis

doomedraven:
btw if you don't want to use poetry, you have requierements.txt :P

VVelox:
> btw if you don't want to use poetry, you have requierements.txt :P

Yeah, but that with the very heavy customization and use of stuff outside of package management makes breakages a common issue. I've not had CAPE work smoothly outside of poetry since our move to it.

Really makes me wish bhyve was had features aimed at actually faking drek so sane package management would be actually possible. Apt/dpkg/ubuntu/debian(and realistically every other Linux distro and common Linux package manager) really leaves a lot to be desired in that area with how out of date and how much of a PITA basic package maintenance is compared to pkgsrc and Ports.

VVelox:
> annoying why? sorry guys didn't have time recently for this. this week is still pretty busy, i guess removing the shebangs should work, but in general if you with poetry you just need to run with `poetry run python submit.py` and if you don't use poetry you just use it as `python3 submit.py`. so i don't know what is the issue there, i will need to test it more in deep, but i would say nobody of us using submit.py as we do not work on server side to add analysis

Honestly this is how I largely add stuff. Honestly I really wish the settings and the like in the web were better exposed via submit.

While as it works currently is not bad for scripting purposes, it's terrible for general usage purposes.

VVelox:
Actually was just thinking what may work better and solve every one's needs.

A bin directory the includes wrappers for common items, queries a config that tells it to use either poetry or python3, and invokes it as needed.

doomedraven:
we just don't use submit is why we don't care about it :P 
about wrapper is not bad idea. but still i don't see a problem to just add few extra words before python. If you want to write the wrapper you are more than welcome. But then that part should also get docs

VVelox:
> we just don't use submit is why we don't care about it :P about wrapper is not bad idea. but still i don't see a problem to just add few extra words before python. If you want to write the wrapper you are more than welcome. But then that part should also get docs

Aye! Will be doing so. Going to re-factor this PR into that. :)

That said as to why, let me ask you, do you less everything and then point the appropriate interpreter at it? I am guessing not.

Having to manually specify a interpreter or the like is a really backwards thing in this day and age. Especially since the install scripts for this install it in a very specific manner that by no means should requires operating in such a backwards manner.

doomedraven:
hm i don't really have answer to that. Would need some brainstorming but i have to finish my dayjob stuff and leave for PTO. I guess is setup related. as for me that isn't a problem to have in my scripts `poetry run python X`. But btw if you don't want use poetry you have requirements.txt so you just need to edit systemd scripts if you installed with cape2.sh to remove `poetry run`. That is the beauty of opensource that you can personalize it as much as you want

VVelox:
> hm i don't really have answer to that. Would need some brainstorming but i have to finish my dayjob stuff and leave for PTO. I guess is setup related. as for me that isn't a problem to have in my scripts `poetry run python X`. But btw if you don't want use poetry you have requirements.txt so you just need to edit systemd scripts if you installed with cape2.sh to remove `poetry run`. That is the beauty of opensource that you can personalize it as much as you want

Ohh! I know and you can definitely expect this PR to be updated with the required bits and docs for this. :)

I am not really expecting an fix for this from you here as I've stepped up and volunteered to get something in place for it. :)

Have fun with the PTO, dude!

doomedraven:
I know, it just good to have some brainstorming, you know few brains sometimes can get some neat ideas than just one ;) Thank you

VVelox:
That said looks like this annoyance has become a non issue for me as I've recently started work on something that calls it, among other things. =^.^=

Basically was wanting some nicer CLI tools in general for CAPE so started working on it.

https://github.com/VVelox/CAPE-Utils

PR #1017, opened 2022-07-20
Processing Improvements

cccs-kevin:
- Configurable behaviour sub-modules
- Bugfix for `add_error` call
- Updating behaviour signature processing to be in-line with how Cuckoo does it
- Add `web_reporting` config to allow both mongodb and elasticsearch reporting to be disabled. This is useful for when you want the REST API that Django supplies, but you don't want the GUI.
- Adding `rtmp` default value
- Toggle the use of `list(process["calls"])` since it is only useful for certain reporting modules, like mongodb.
- Adding check if file exists for PE-parsing
- Adding a `finally` statement in the scheduler from Cuckoo that marks tasks as completed if there is an AnalysisManager error.
- Fixing ramboost bug
- Adding `matched` attribute to the `Signature` class to avoid duplication of signature hits
- Adding check for if sample doesn't exists in db

Something that I have noticed in CAPE which I don't think(?) was an issue in Cuckoo, is the length of time that it takes to iterate through the `calls` of a process here https://github.com/kevoreilly/CAPEv2/blob/1822c2ed8700548878dbedd6c676bb403e478961/lib/cuckoo/core/plugins.py#L494

For instance, I have a sample (a batch file with one line `start "" https://www.stackoverflow.com`) that creates a process which opens `iexplore.exe`. This process makes a lot of calls. Without `ramboost` turned on, the processing of the Bson logs in the `behavior.py` module takes ~2 seconds, and then the iteration of calls when applying evented signatures takes ~30 seconds (even with my improvements in this PR). When I turn `ramboost` on, the processing of the Bson logs takes ~30 seconds and the signatures take ~2 seconds. So basically it always takes 32 seconds, which _feels_ like that is too long to just parse/iterate through logs. Is this issue something that people are aware of / can duplicate?

### DIFF
--- a/conf/processing.conf
+++ b/conf/processing.conf
@@ -57,6 +57,12 @@ enabled = no
 
 [behavior]
 enabled = yes
+# Toggle specific modules within the BehaviorAnalysis class
+anomaly = yes
+processtree = yes
+summary = yes
+enhanced = yes
+encryptedbuffers = yes
 
 [debug]
 enabled = yes
--- a/conf/reporting.conf
+++ b/conf/reporting.conf
@@ -230,3 +230,6 @@ command=/foo/bar.pl
 # run statistics, this may take more times.
 [runstatistics]
 enabled = no
+
+[malheur]
+enabled = no
--- a/conf/web.conf
+++ b/conf/web.conf
@@ -240,3 +240,6 @@ enabled = yes
 
 [during_script]
 enabled = yes
+
+[web_reporting]
+enabled = yes
--- a/lib/cuckoo/common/abstracts.py
+++ b/lib/cuckoo/common/abstracts.py
@@ -756,6 +756,7 @@ class Signature:
         self._current_call_raw_dict = None
         self.hostname2ips = {}
         self.machinery_conf = machinery_conf
+        self.matched = False
 
     def statistics_custom(self, pretime, extracted: bool = False):
         """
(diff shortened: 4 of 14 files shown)


doomedraven:
btw let me know once is ready to merge

cccs-kevin:
> btw let me know once is ready to merge

Sure thing, there are still a few more things I want to do

doomedraven:
Question about try excepto of file not found, whould t it be better to just use if os path exist?

cccs-kevin:
> Question about try excepto of file not found, whould t it be better to just use if os path exist?

Yes, I'll update

doomedraven:
Thanks

doomedraven:
amazing work Kevin. thank you.

About ram_boost, that is feature from cuckoo-modified i think, maybe cuckoo has it, it already ages since i last checked cuckoo code. I have that off on all my servers. I will create issue for this to review once i finish huge RE  that i have on my table

doomedraven:
something is broken, it says all my vms are dead after merge, investigating

doomedraven:
ok nevermind, for some reason qemu disappeared from my server %)

cccs-kevin:
> amazing work Kevin. thank you.
> 
> About ram_boost, that is feature from cuckoo-modified i think, maybe cuckoo has it, it already ages since i last checked cuckoo code. I have that off on all my servers. I will create issue for this to review once i finish huge RE that i have on my table

The issue that I'm talking about isn't even really about `ramboost`, it's about how long it takes to iterate through the ParseProcessLog object https://github.com/kevoreilly/CAPEv2/blob/200a80818f82f5b9487e9242f8da32ae969aa27c/modules/processing/behavior.py#L38

doomedraven:
once i will have some spare time i will try to investigate it i have created issue to not forget about it https://github.com/kevoreilly/CAPEv2/issues/1024

PR #1018, opened 2022-07-20
resubmit update and simplify code

doomedraven:


doomedraven:
hm something is broken, trying to spot what, it says that my VMs are dead

PR #1027, opened 2022-07-26
Use runserver_plus for web so that we can see exceptions

cccs-kevin:
We are importing the `django-extensions` package, which provides `runserver_plus`. This module allows us to see the stacktraces for REST exceptions, which are incredibly useful compared to not having them.

The --keep-meta-shutdown is to address this bug for older versions of Python3 https://stackoverflow.com/questions/72166259/werkzeug-server-is-shutting-down-in-django-application


doomedraven:
ooo that is interesting, thank you

PR #1031, opened 2022-07-29
small fix

enzok:
i had to make this change to for URLs to submit on my instance, not sure if there are actually supposed to be 3 'list_of_files'


doomedraven:
almost, https://github.com/kevoreilly/CAPEv2/blob/3caff511f046359cffcd70f18a2450567bef89a5/web/submission/views.py#L269-L275

see there is 2 different cases, it should be 3 not 2 as the rest uses 3, fixing

PR #1032, opened 2022-08-01
fix: duplicate signatures under same TTP ID

TheMythologist:
Fixes the following bug:
brave_pn2sj3RcCl

Apologies for not fixing this bug earlier, my deployment of CAPE did not encounter this error.


doomedraven:
thank you

PR #1035, opened 2022-08-01
Fix regression in VM selection.

tbeadle:
https://github.com/kevoreilly/CAPEv2/pull/1020 introduced a regression
in selection of VM's for x86 tasks where it would no longer allow
selection of an x64 VM. This reintroduces that ability so that x86 tasks
will select x86 OR x64 machines, preferring x86 over x64.


kevoreilly:
Thank you

cccs-kevin:
Thanks @tbeadle !

PR #1044, opened 2022-08-03
Fix machine conf output error

qux-bbb:
python writelines no line breaks.

### DIFF
--- a/utils/clone-machines.py
+++ b/utils/clone-machines.py
@@ -180,8 +180,10 @@ def print_machines_config(machines):
                 f"tags = x64",
             ]
 
-            machines_file.writelines(machine_lines)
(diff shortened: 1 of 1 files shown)


doomedraven:
im back, could you post the error and new output example? without that i don't know what to expect and how it was. so please in future provide more detailed before/after that will speedup stuff

qux-bbb:
Just like I said, python writelines no line breaks.  
```python
names = ["Alice", "Bob", "Cindy"]

the_file = open("before.txt", "w")
the_file.writelines(names)
the_file.close()

the_file = open("after.txt", "w")
the_file.write("\n".join(names))
the_file.close()
```
writelines

doomedraven:
thank you

PR #1054, opened 2022-08-07
Updated docs

RazviOverflow:
- [Added instructions for setting a Static IP in Win10](https://github.com/RazviOverflow/CAPEv2/commit/3b44499c7852eb7d0ab1dd291b8bbceef273055d)

- [Supressed underlining too short](https://github.com/RazviOverflow/CAPEv2/commit/90aaaaad68f7e9af21bdd006debb387284300561)

- [Added instructions to disable firewall and automatic update on Win 10](https://github.com/RazviOverflow/CAPEv2/commit/fd2cb3e46618a42bb2a3553ff3d5be437d82d410)

- [Added documentation for guest's Python installation](https://github.com/RazviOverflow/CAPEv2/commit/ba7e9177816e3124b0bba8cc5c85d888fd46d7ab)


doomedraven:
thank you a lot for this @RazviOverflow. i have only one last point, the 192.168.122.0/24. the 122 range is default KVM range, so that can be used as ANTIVM check. Can you add small description for that. Also 122 is __NAT__. User should create extra network interface with range that is not default hypervisor range. Can you please add it?

doomedraven:
i merge this so docs can update till you add another point from previous msg

RazviOverflow:
You are most welcome @doomedraven! Glad to help. 

Regarding the 122 range, I will create a **warning**  informing that it can be detected and I'll add description + images on how to create a new IP range (something similar to what you did back in the day in your post creating the `hostonly` network). 

If there is not NAT, the machine has no internet connection, right? It may be needed. 

I will PR as soon as I fix the docs and add the needed info.

doomedraven:
machine will get internet with rooter.py if route internet selected. In hostonly way it prevents from traffic leaks, connecting to other machines. Basically there is no choice for more secure setup.

Thank you

doomedraven:
@RazviOverflow FYI https://docs.vmware.com/en/VMware-Workstation-Pro/16.0/com.vmware.ws.using.doc/GUID-D9B0A52D-38A2-45D7-A9EB-987ACE77F93C.html

PR #1084, opened 2022-08-16
Fix URL submission

enzok:
proposed solution for url submissions

### DIFF
--- a/lib/cuckoo/core/database.py
+++ b/lib/cuckoo/core/database.py
@@ -809,8 +809,16 @@ class Database(object, metaclass=Singleton):
         @return: boolean indicating if a relevant machine is available
         """
         # Are there available machines that match up with a task?
-        task_arch = next((tag.name for tag in task.tags if tag.name in ["x86", "x64"]), "")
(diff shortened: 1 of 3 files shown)


lib/cuckoo/core/database.py
```diff
@@ -809,8 +809,16 @@ def is_relevant_machine_available(self, task: Task) -> bool:
...
         """
         # Are there available machines that match up with a task?
-        task_arch = next((tag.name for tag in task.tags if tag.name in ["x86", "x64"]), "")
-        task_tags = [tag.name for tag in task.tags if tag.name != task_arch]
+        task_tags = []
+        for tag in task.tags:
+            if tag.name in ["x86", "x64"]:
+                task_arch = tag.name
+            else:
+                task_tags.append(tag.name)
```
enzok, line 817: this is a list of non-arch tags to match against machine tags (which shouldn't have arch in them anymore!)

enzok:
Marks task as running but doesn't start task

enzok:
This solution works for url submissions which currently require a machine to be selected at submission.

PR #1085, opened 2022-08-16
Use lists for task archs so that files and urls can have archs and tags

cccs-kevin:
I believe this PR solves the issue that https://github.com/kevoreilly/CAPEv2/pull/1084 is trying to solve at its root.

### DIFF
--- a/lib/cuckoo/core/database.py
+++ b/lib/cuckoo/core/database.py
@@ -957,7 +957,7 @@ class Database(object, metaclass=Singleton):
         Allow x64 machines to be returned when requesting x86.
         """
         if arch:
-            if arch == ["x86"] or set(arch) == {"x64", "x86"}:
+            if "x86" in arch:
                 # Prefer x86 machines over x64 if x86 is what was requested.
                 machines = machines.filter(Machine.arch.in_(("x64", "x86"))).order_by(Machine.arch.desc())
             else:
--- a/lib/cuckoo/core/database.py
+++ b/lib/cuckoo/core/database.py
@@ -809,10 +809,10 @@ class Database(object, metaclass=Singleton):
         @return: boolean indicating if a relevant machine is available
         """
         # Are there available machines that match up with a task?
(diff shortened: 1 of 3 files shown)


lib/cuckoo/core/database.py
```diff
@@ -957,7 +957,7 @@ def filter_machines_by_arch(machines, arch):
         Allow x64 machines to be returned when requesting x86.
         """
         if arch:
-            if arch == "x86":
+            if arch == ["x86"] or set(arch) == {"x64", "x86"}:
```
tbeadle, line 960: can we just use `if "x86" in arch:`?

cccs-kevin: brilliant, see https://github.com/kevoreilly/CAPEv2/pull/1085/commits/a74f383bd09ef41d9ae9a03839e6658813365967

lib/cuckoo/core/database.py
```diff
@@ -957,7 +957,7 @@ def filter_machines_by_arch(machines, arch):
         Allow x64 machines to be returned when requesting x86.
         """
         if arch:
-            if arch == "x86":
+            if "x86" in arch:
                 # Prefer x86 machines over x64 if x86 is what was requested.
                 machines = machines.filter(Machine.arch.in_(("x64", "x86"))).order_by(Machine.arch.desc())
             else:
```
tbeadle, line 963: question: Does this mean that, if the task has no arch tags, then it wouldn't find a machine at all because it would be looking for machines whose architecture is IN an empty list?

doomedraven: hm it always adds x86 tag if no other matches any criteria

cccs-kevin: It is possible for a task to have no arch tag if you have both of these config options switched off
https://github.com/kevoreilly/CAPEv2/blob/master/conf/web.conf#L41:L44

I have both switched off in my environment and tasks are able to get assigned.

PR #1093, opened 2022-08-17
if stap enabled, self.stop should be called to finish and upload log

micahflack:


micahflack (commented):
if stap is enabled it should call self.stop at the end of self.run and upload the stap.log; it is not called anywhere else to finish the STAP module

otherwise it would be uploaded by FileCollector - but it is disabled there as well

analyzer/linux/modules/auxiliary/stap.py
```diff
@@ -70,6 +70,9 @@ def has_stap(p):
 
         stap_stop = timeit.default_timer()
         log.info("STAP aux module startup took %.2f seconds", stap_stop - stap_start)
+
+        self.stop()
```
cccs-kevin, line 74: is this not called by https://github.com/kevoreilly/CAPEv2/blob/master/analyzer/windows/analyzer.py#L691?

micahflack: oh, yeah - you're right. I definitely overlooked the aux.stop() at line 322; thought it wasn't called because I didn't see it after aux.start() @ 203.

PR #1094, opened 2022-08-17
changed STAP.py to call self.stop to finish module and upload log

micahflack:
the STAP module should call self.stop() before finishing self.run() to ensure the process is terminated and to upload the stap.log. If this is not done, the stap.log will not be collected because it is ignored under the FileCollector module.


micahflack:
@cccs-kevin tagging because you might be interested

PR #1095, opened 2022-08-17
Revert "changed STAP.py to call self.stop to finish module and upload log"

micahflack:
Reverts kevoreilly/CAPEv2#1094; cccs-kevin corrected my PR, the aux.stop() was already called later in analyzer.py


doomedraven:
0 files changed

PR #1102, opened 2022-08-21
Enable packet forwarding for IPv4

qux-bbb:
Always need to enable ip forwarding, so put it in the installation script.

### DIFF
--- a/installer/cape2.sh
+++ b/installer/cape2.sh
@@ -813,6 +813,11 @@ EOF
         echo "net.bridge.bridge-nf-call-arptables = 0";
     } >> /etc/sysctl.conf
 
+    # enable packet forwarding for IPv4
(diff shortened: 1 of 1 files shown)


doomedraven:
never need to enable it. rooter does that for you. why just not to use it?

qux-bbb:
If you do not enable IP forwarding, rooter is invalid.  
In documentation:  
https://capev2.readthedocs.io/en/latest/installation/host/routing.html#using-per-analysis-network-routing  
```r
Both global routing and per-analysis routing require ip forwarding to be enabled:
echo 1 | sudo tee -a /proc/sys/net/ipv4/ip_forward
```

What I did was add it to the installation script.

doomedraven:
ok you are right, i confused with another option, my bad

qux-bbb:
My bad description. Thank you for the merge.

PR #1107, opened 2022-08-23
Updating scheduler/database to improve the time a machine is locked for

cccs-kevin:
Something that I've noticed in the following environment:
- you have ~100 clients that submit tasks and poll until analysis completion,
- you have 100 machines that are eligible for all tasks
- each analysis runs for ~2 minutes

The logic in the `scheduler.py` with the `machine_lock` prevents 100 machines to be assigned a task, and reaches an equilibrium of ~50 tasks in the pending queue with ~30-50 machines in use. Here is why:
- It takes ~2 seconds for a task to be assigned to a machine, and the majority of this time is spent waiting to acquire the `machine_lock` https://github.com/kevoreilly/CAPEv2/blob/master/lib/cuckoo/core/scheduler.py#L184
    - Therefore every ten seconds, ~5 tasks are *always* assigned to a machine
- Analyses complete periodically, depending on the file type
    - Every ten seconds, >5 tasks are completed and machines are freed up

My proposition in this PR:
- Speed up the time that a `machine_lock` is acquired for by moving stuff that isn't required outside of the time when the `machine_lock` is locked.
    - Moving the `route_network` method does not need to be in this section and takes a bit of time, so that cuts the `machine_lock` blocking time.
    - Merging the `guest_start` and `set_task_vm` methods into one since both methods commit & refresh the DB for a Task row. This speeds up the blocking time a bit.
- Remove the time to sleep time between scheduler loops. This results in more DB calls, but removes an arbitrary sleep in the loop. There may be a way here to avoid sleeping in a smarter way? Any input is welcome.

Since the `availables()` method is called every iteration of the scheduler loop, the `<query>.count()` method in SQLAlchemy is known to be slow, so here is a thread talking about a way to speed it up: https://gist.github.com/hest/8798884

These modifications help a bit, sometimes 12 tasks are assigned within a 10 second period... which is great BUT I think there is a way to improve the use of the `machine_lock` usage. Maybe with semaphores? I'm not sure...

Maybe something smart like a bounded semaphore that can be acquired X times where X is the number of relevant available machines?

### DIFF
--- a/lib/cuckoo/core/database.py
+++ b/lib/cuckoo/core/database.py
@@ -781,25 +781,33 @@ class Database(object, metaclass=Singleton):
             session.close()
 
     @classlock
-    def set_task_vm(self, task_id, vmname, vm_id):
-        """Set task status.
+    def set_task_vm_and_guest_start(self, task_id, vmname, vmlabel, vm_id, manager):
+        """Set task status and logs guest start.
         @param task_id: task identifier
         @param vmname: virtual vm name
-        @return: operation status
+        @param label: vm label
+        @param manager: vm manager
+        @return: guest row id
         """
         session = self.Session()
+        guest = Guest(vmname, vmlabel, manager)
         try:
+            guest.status = "init"
             row = session.query(Task).get(task_id)
 
             if not row:
                 return
 
+            row.guest = guest
             row.machine = vmname
             row.machine_id = vm_id
             session.commit()
+            session.refresh(guest)
+            return guest.id
         except SQLAlchemyError as e:
-            log.debug("Database error setting status: %s", e)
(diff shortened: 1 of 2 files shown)


cccs-kevin:
Closed by https://github.com/kevoreilly/CAPEv2/pull/1757

doomedraven:
Thanks, great explanation and improvement

doomedraven:
i just read the linked gist, and i saw that with `with_only_columns` variation in past i was able to make dist.py way much faster, so i slowly started to move toward that, as it only gets desired columns instead of everything what is useless, slow and takes a lot of ram, so will have to get this code from gist into core, thank you

kevoreilly:
Wow this is awesome - thank you ❤️

PR #1113, opened 2022-08-25
ensure JsonDump run after the SubmitCAPE

qux-bbb:
When I submit a file, it triggered automatically submitting sub task, but I can not find "CAPE_children" in its report.json.  
The reason is JsonDump run before SubmitCAPE.  
Add order can fix it.


doomedraven:
thanks

PR #1120, opened 2022-08-30
Setting a default package if sflock throws an exception when identifying a file

cccs-kevin:
If sflock fails to identify something because of a 3rd party package exception, the REST API returns a 500. But we should at least TRY to run this sample rather than not add it to the database at all.

### DIFF
--- a/lib/cuckoo/core/database.py
+++ b/lib/cuckoo/core/database.py
@@ -1595,7 +1595,13 @@ class Database(object, metaclass=Singleton):
             if not config and not only_extraction:
                 if not package:
                     f = SflockFile.from_path(file)
-                    tmp_package = sflock_identify(f)
+
+                    try:
+                        tmp_package = sflock_identify(f)
+                    except Exception as e:
+                        log.error(f"Failed to sflock_ident due to {e}")
+                        tmp_package = "cmd"
+
                     if tmp_package and tmp_package in sandbox_packages:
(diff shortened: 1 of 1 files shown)


doomedraven:
Hello thanks, im fine to merge this, but as side i would love to know what problem you got, you can share that in PM too

cccs-kevin:
This sample https://www.virustotal.com/gui/file/b6a71449958283718f01e6be9f7a9eb057479072c1cc55f26a11b033ea210271

when submitted to CUCKOO, not CAPE, extracts a dropped DOS file with the SHA256 of eaedceea8b2fc89872bd5e4c4a6f90721319b4fdb23deccff4630c7087b5de4e. This file when sent to `sflock_ident` raises ` pefile.PEFormatError: 'Invalid e_lfanew value, probably not a PE file'` from the `pefile` module which is called here `File "sflock/ident.py", line 486, in identify  pe = pefile.PE(data=f.contents, fast_load=True)` and this exception is unhandled, causing the REST API return status code 500.

Here is the DOS file:

[eaedceea8b2fc89872bd5e4c4a6f90721319b4fdb23deccff4630c7087b5de4e.zip](https://github.com/kevoreilly/CAPEv2/files/9463708/eaedceea8b2fc89872bd5e4c4a6f90721319b4fdb23deccff4630c7087b5de4e.zip)

PR #1126, opened 2022-08-31
Adding the ability to extract an archive (ISO, VHD, etc) into a folder and run an executable

cccs-kevin:
Tested with a few ISO and VHD files and it works pretty well. You need 7-zip installed on your guest machines.

### DIFF
--- a/analyzer/windows/modules/packages/archive.py
+++ b/analyzer/windows/modules/packages/archive.py
@@ -1,0 +1,163 @@
+# Copyright (C) 2010-2015 Cuckoo Foundation.
+# This file is part of Cuckoo Sandbox - http://www.cuckoosandbox.org
+# See the file 'docs/LICENSE' for copying permission.
+
+import logging
+import os
+import shutil
+import subprocess
+
+try:
+    import re2 as re
+except ImportError:
+    import re
+
+from lib.common.abstracts import Package
+from lib.common.common import check_file_extension
+from lib.common.exceptions import CuckooPackageError
+
+log = logging.getLogger(__name__)
+
+
+FILE_NAME_REGEX = re.compile("[\s]{2}([a-zA-Z0-9\.\-_\\\\]+)\\r")
+EXE_REGEX = re.compile(r"(\.exe|\.dll|\.scr|\.msi|\.bat|\.lnk|\.js|\.jse|\.vbs|\.vbe|\.wsf|\.ps1)$", flags=re.IGNORECASE)
+
+class Archive(Package):
+    """Archive analysis package."""
+
+    PATHS = [
+        ("SystemRoot", "system32", "cmd.exe"),
+        ("SystemRoot", "system32", "wscript.exe"),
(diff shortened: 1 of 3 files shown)


doomedraven:
there was in past zip.py that was doing the same, i personally prefer to move that to sflock as otherwise you lose relationship  between archive and childs

cccs-kevin:
> there was in past zip.py that was doing the same, i personally prefer to move that to sflock as otherwise you lose relationship between archive and childs

In what way is the relationship lost? The archive file is uploaded and a new folder is created with the same base name of the archive, and the children are extracted into that folder.

kevoreilly:
Thanks for this PR.

I just tried it but unfortunately it crashed:

```
2022-09-06 14:02:24,763 [root] INFO: You probably submitted the job with wrong package
  File "C:\tmpljosyc8h\analyzer.py", line 528, in run
    pids = self.package.start(self.target)
  File "C:\tmpljosyc8h\modules\packages\archive.py", line 139, in start
    os.mkdir(root)
FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'C:\\Users\\Rebecca\\AppData\\Local\\Temp\\1fc7b0e1054d54ce8f1de0cc'

The above exception was the direct cause of the following exception:

  File "C:\tmpljosyc8h\analyzer.py", line 1397, in <module>
    success = analyzer.run()
  File "C:\tmpljosyc8h\analyzer.py", line 534, in run
    raise CuckooError(f'The package "{package_name}" start function encountered an unhandled exception: {e}') from e
lib.common.exceptions.CuckooError: The package "modules.packages.archive" start function encountered an unhandled exception: [WinError 183] Cannot create a file when that file already exists: 'C:\\Users\\Rebecca\\AppData\\Local\\Temp\\1fc7b0e1054d54ce8f1de0cc'
```

Hopefully simple to fix.

cccs-kevin:
I'll test it out and report back

cccs-kevin:
Works fine for me... Are you using the most updated [`archive` package](https://github.com/kevoreilly/CAPEv2/blob/master/analyzer/windows/modules/packages/archive.py)? Also question, is your submitted sample named `1fc7b0e1054d54ce8f1de0cc95976081c7a85c7926c03172a3ddaa672690042c` or is it renamed to `1fc7b0e1054d54ce8f1de0cc95976081c7a85c7926c03172a3ddaa672690042c.iso`?

kevoreilly:
It would be nice to have the possibility of putting 7z.exe in analyzer/windows/bin instead of needing to have it baked into the vm image...

cccs-kevin:
> I am using the most up-to-date package, but did not rename the iso - if I do I see this is no longer an issue.

Let me know if [this PR](https://github.com/kevoreilly/CAPEv2/pull/1132) works without the renaming.

cccs-kevin:
> It would be nice to have the possibility of putting 7z.exe in analyzer/windows/bin instead of needing to have it baked into the vm image...

https://github.com/kevoreilly/CAPEv2/pull/1133

PR #1132, opened 2022-09-06
Handle case where file name does not have an extension

cccs-kevin:
To address https://github.com/kevoreilly/CAPEv2/pull/1126#issuecomment-1238237401


kevoreilly:
Thanks

PR #1134, opened 2022-09-06
Alternate dump method for msi extraction

enzok:
There are some msi files that aren't handle by msi-extract


doomedraven:
did small cleanup

PR #1136, opened 2022-09-07
fix suricata-update in crontab line added each time cape2.sh is executed

LeanVel:
Grep is interpreting the "*" and it is not finding the line when it is already in the crontab. Adding -F flag to grep will avoid adding the suricata-update like multiple times to crontab. 

-F, --fixed-strings
          Interpret  PATTERN  as  a  list  of  fixed strings, separated by
          newlines, any of which is to be matched.  (-F  is  specified  by
          POSIX.)


doomedraven:
thanks, specially for context

PR #1140, opened 2022-09-08
Update cape2.sh

LeanVel:
- Addressed the apt-key deprecation following the apt-key (8) recommendations
- Changed back the repo of mongo to use $(lsb_release -cs) instead of hard coding distro. https://repo.mongodb.org/apt/ubuntu/dists/jammy/mongodb-org/6.0/multiverse/ does exist now
- Addressed a temporary issue compiling yara-python
- Addressed an issue with the useage of poetry and the usage of keyrings


doomedraven:
thanks

PR #1159, opened 2022-09-21
Fix poetry path and guacamole asgi app name

rgsax:


doomedraven:
the path is looks like is depended of each OS, so we need to use SED i guess, but the other fix is correct, thank you

PR #1165, opened 2022-09-24
Create Trojan_Rozena.yar

sqrtZeroKnowledge:


doomedraven:
Hello, thank you for you contribution. The community yara should go to that repo https://github.com/kevoreilly/community. if you want i can move it. or if you want you can move it, let me know

kevoreilly:
Also yara is not in keeping with cape standard.

'Trojan' should not be part of the family name, it should just be 'Rozena'.

Classification in cape needs a cape_type which here would be "Rozena Payload".

There is only a single pattern which is poor for resisting FP and future breakages. A minimum of two but preferably three patterns would be better.

A hash should be included with the PR or better in the yara sig metadata so we can test the yara and know from which sample(s) it was derived.

PR #1196, opened 2022-10-11
Adding note about archive usage

cccs-kevin:
I've made a few talks recently regarding the increased presence of non-zip archive files being used by malicious actors in present day. My work on the `archive.py` package provides CAPE's best chance of detecting these samples and network IOCs. This note is to assist those who are using a client to submit files to CAPE that is not [Assemblyline ](https://cyber.gc.ca/en/tools-services/assemblyline).

### DIFF
--- a/docs/book/src/usage/packages.rst
+++ b/docs/book/src/usage/packages.rst
@@ -28,6 +28,9 @@ The following is a list of the existing packages in alphabetical order:
 
     * ``archive``: used to run and analyze **archives such as ISO, VHD and anything else that 7-Zip can extract**.
 
+        *NB*: Passing ``file=`` as a task option will ensure that the entire archive is passed to the victim VM and extracted there, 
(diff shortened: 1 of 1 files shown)


doomedraven:
thank you for your work on it

cccs-kevin:
For those interested, here is one of such talks: https://www.youtube.com/watch?v=-70Mlkmtdds&t=11056s

doomedraven:
nice talk, i did quick look skipping some parts. but i think your explanation of archive package would be probably even nice to link to docs so newcomers could use that. what do you think?

doomedraven:
;) https://github.com/kevoreilly/CAPEv2/commit/54083aa71543c0923ea6f96102bb7ff3157d8ed4

doomedraven:
linked in docs too https://capev2.readthedocs.io/en/latest/usage/packages.html

PR #1208, opened 2022-10-20
Miscellaneous changes

ClaudioWayne:
- No virt-manager install in cape.sh, so removed in docs.
- Fix TOR sourcelist entry
- Match libvirt-python version with installed libvirt version
- Fix guacamole fstab entry
- Fix guacamole recording path in web.conf
- Guacamole docs hint for web server without HTTPS config


doomedraven:
im not agree with half of the changes. i will edit them once i have some spare time

PR #1212, opened 2022-10-21
Fix elasticsearch document mapping errors

CorraMatte:
Bug fixing for Elasticserch reporting.
It now ingests correctly reports created by file analyses.


doomedraven:
thank you

PR #1228, opened 2022-11-01
Fix default proxy value

qux-bbb:
If proxy does not have a default value, the script will show this error:  
`TypeError: flare_capa() missing 1 required positional argument: 'proxy'`


kevoreilly:
Thank you

PR #1257, opened 2022-11-25
Wrapping the reimage and delete threads in a try-catch

cccs-kevin:
We don't want these threads to crash, ever!


kevoreilly:
Thank you 🙏

Issue #1182, opened 2022-10-04
Add documentation for additional analysis packages


cccs-kevin:
## About accounts on [capesandbox.com](https://capesandbox.com/)
* Issues isn't the way to ask for account acctivation. Ping capesandbox in [Twitter](https://twitter.com/capesandbox) with your username

## This is opensource and you getting __free__ support so be friendly!
* Free support from doomedraven ended, no whiskey no support. For something he updated the documentation :)

# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest version
- [x] I did read the README!
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I'm reporting the issue to the correct repository (for multi-repository projects)
- [x] I'm have read all configs with all optional parts

# Expected Behavior

Please describe the behavior you are expecting. __If your samples(x64) stuck in pending ensure that you set tags=x64 in hypervisor conf for x64 vms__

I would like to see all available analysis packages represented in the [documentation](https://capev2.readthedocs.io/en/latest/usage/packages.html).

# Current Behavior

What is the current behavior?

Only a subset of all available analysis packages are listed. For example, these are the currently available analysis packages in Windows:

```
Shellcode-Unpacker.py
Shellcode.py
Shellcode_x64.py
UPX.py
UPX_dll.py
Unpacker.py
Unpacker_dll.py
Unpacker_js.py
Unpacker_ps1.py
Unpacker_regsvr.py
Unpacker_zip.py
access.py
applet.py
archive.py
chm.py
chrome.py
cpl.py
dll.py
doc.py
doc2016.py
edge.py
eml.py
exe.py
firefox.py
generic.py
hta.py
html.py
hwp.py
ichitaro.py
ie.py
inp.py
jar.py
js.py
js_antivm.py
lnk.py
mht.py
msbuild.py
msg.py
msi.py
nsis.py
ollydbg.py
pdf.py
ppt.py
ppt2016.py
processes.py
processes_simple.py
ps1.py
ps1_x64.py
pub.py
pub2016.py
python.py
rar.py
reg.py
regsvr.py
sct.py
service.py
service_dll.py
swf.py
vawtrak.py
vbejse.py
vbs.py
wsf.py
xls.py
xls2016.py
xlst.py
xps.py
zip.py
zip_compound.py
```

# Failure Information (for bugs)

[...]

PR #1258, opened 2022-11-25
Updated the packages documentation

cccs-kevin:
Closes https://github.com/kevoreilly/CAPEv2/issues/1182


kevoreilly:
Thanks! But is it intended to duplicate, for example, `procmemdump` so many times?

cccs-kevin:
> Thanks! But is it intended to duplicate, for example, `procmemdump` so many times?

Good question, it seems like `procmemdump`, `free` and `dll` were options for every analysis package prior to this change. So I just kept with that theme. I can include these options at the top of the doc with a blurb and remove them from the rest of the doc?

PR #1259, opened 2022-11-25
Refactoring 'processes' package

cccs-kevin:
- `processes_simple` and `processes` were pretty much the sample thing in theory, other than the commented out code. So I removed `processes_simple`.
- The `processes` package essentially was the winword equivalent of the `js_antivm` package, so I renamed it to `doc_antivm`.
- There were some applications that are not installed at default or had the incorrect path, so I removed their execution during startup.
- Updating the comments in the `js_antivm` package, as well as only creating 20 calcs vs the original 50 since 50 is overkill.


kevoreilly (approved):
Looks good to me

PR #1263, opened 2022-11-30
Path() expects a str or os.PathLike object

enzok:


cccs-kevin (changes requested):
Hey @enzok, thanks for the PR! Can you please add a set of unit tests for this method https://github.com/kevoreilly/CAPEv2/blob/master/tests/test_demux.py that test prior to the change and after the change?

PR #1271, opened 2022-12-08
Only one name will be displayed for the same malware family

fire-flying:
Only one name will be displayed for the same malware family
before
image
image

after 
image
image

### DIFF
--- a/modules/signatures/CAPE.py
+++ b/modules/signatures/CAPE.py
@@ -610,6 +610,29 @@ class CAPEDetectedThreat(Signature):
     minimum = "1.3"
     evented = True
 
+    _family_map = {"Cobalt Strike": ["CobaltStrikeBeacon", "Meterpreter", "Cobalt"]}
+
+    def alias_to_family(self):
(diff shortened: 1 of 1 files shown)


kevoreilly:
I'm not a fan of this approach - it seems to me it's just masking the underlying problem of having multiple detection names for the same family. Why is there a 'Cobalt' detection and why can't it be fixed to be 'CobaltStrikeBeacon'? And if Meterpreter is not CobaltStrike, why call it that in the detection and fix the incorrect name?

fire-flying:
> I'm not a fan of this approach - it seems to me it's just masking the underlying problem of having multiple detection names for the same family. Why is there a 'Cobalt' detection and why can't it be fixed to be 'CobaltStrikeBeacon'? And if Meterpreter is not CobaltStrike, why call it that in the detection and fix the incorrect name?

This is just an example and can be configured according to your own habits. What I want to solve is the problem of malicious files of the same family showing Multiple on the nearest page

PR #1285, opened 2022-12-19
Bump libvirt-python version to 8.10.0

orenyomtov:
To keep it the same version as 
https://github.com/doomedraven/Tools/blob/master/Virtualization/kvm-qemu.sh#L65 as mentioned in
https://github.com/kevoreilly/CAPEv2/pull/952#issuecomment-1163430794

Fixes the following error:
```
Missing type converters:
virTypedParameterPtr:2
ERROR: failed virDomainRestoreParams
ERROR: failed virDomainSaveParams
```


doomedraven:
thanks

PR #1287, opened 2022-12-21
Fix file copy

enzok:


doomedraven:
@cccs-kevin ?

cccs-kevin:
Wow @enzok , so you're saying that if the archive had a flat structure, that those files weren't being copied to C:?

cccs-kevin (approved):

kevoreilly:
Thanks all 🙏

cccs-kevin:
@enzok is 🎅

enzok:
> Wow @enzok , so you're saying that if the archive had a flat structure, that those files weren't being copied to C:?

Yes the root files were excluded. Only sub dirs where copied.

PR #1292, opened 2023-01-01
Clean cape_name when have exception or yara rule do not have cape_type

vest12385:
UPX rule do not have cape_type, so if sample packed by UPX, will mess up config extract result.

Test sample : d607e4cf6254db79c5a50a6e59c25f9387951f545da0b1475719209d8d357f64


doomedraven:
thank you, i did small change

vest12385:
Sure, Thanks for maintenance this amazing project.

PR #1296, opened 2023-01-03
malscore config in search results

enzok:


web/templates/analysis/search.html
```diff
@@ -312,7 +312,9 @@ <h3 class="panel-title">Search Results</h3>
                     {% if config.virustotal %}
                     <th width="5%">VT</th>
                     {% endif %}
+                    % if config.malscore %}
```
doomedraven, line 315: bug here, fixing

PR #1304, opened 2023-01-06
Fix KeyError

enzok:
fix condition where raw_value key may not exist, also raw_value_string was introduced, not sure if this also needs to be checked here, but added just in case


doomedraven:
can you check my changes?

PR #1311, opened 2023-01-12
Fix bugs/speed up/clean up generic file extractors.

tbeadle:
This addresses some issues with how the generic file extractors were being handled and cleans up the code to make them behave a bit more consistently.
 - It was often the case, on a somewhat loaded system, that the first task to be checked for "ready()" status would not be ready and the timeout would not have elapsed. In this case, we would sleep for 5 seconds. For samples where there are many files to process, we would end up waiting these extra 5 seconds for every one. This adds up quickly. The method for handling the child processes is improved so that we don't need to do that.
 - The "took_seconds" calculation was inaccurate because it would be counting from when the process pool was created, not from when we starting trying to run the tool, which could be later based on how many workers are in the pool.
 - Factored out lots of code so that the unique tool-handling functions are limited to their specific duties.
 - Go back to using a ProcessPool but fix how the 'results' and 'duplicated' objects are being handled. They were previously being updated in child processes. These updates would not be reflected in the parent process, essentially losing this data. This was recently addressed by using a ThreadPool instead, which is certainly one way to do it, but this addresses the other issues mentioned above as well and cleans up how the 'results' and 'duplicated' objects are being handled without having to deal with race conditions, which, I believe, are possible in the ThreadPool version.
 - Add some additional logging.

### DIFF
--- a/lib/cuckoo/common/integrations/file_extra_info.py
+++ b/lib/cuckoo/common/integrations/file_extra_info.py
@@ -1,17 +1,20 @@
+import concurrent.futures
+import contextlib
+import functools
 import hashlib
 import json
 import logging
 import os
+import pebble
 import shlex
 import shutil
 import subprocess
 import tempfile
-import time
 import timeit
 from multiprocessing.context import TimeoutError
 from multiprocessing.pool import ThreadPool
 from pathlib import Path
-from typing import List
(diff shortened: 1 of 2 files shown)


doomedraven:
thank you

PR #1312, opened 2023-01-17
make max filename len chomping tweakable

VVelox:
This allows the max filename length as well as the length it will be sanitized to tweakable via config file.


doomedraven:
just spotted, so FYI
`config = Config()` == `cuckoo_cfg = Config("cuckoo")`

PR #1317, opened 2023-01-20
Fix bug memory Analysis Information

goheesheng:
Notice that the Path and Load Times do not have the proper information.

We rectified the HTML templates such that the Path and Load Times would be properly displayed.

memory_analysis_info


doomedraven:
thank you

PR #1329, opened 2023-01-25
Fix pagination bug

goheesheng:
While the first page has the page navigation bar, it disappears on the second page. This was due to missing codes in Python file "views.py" which controlled the information displayed on the HTML templates.

navigationbug

After rectifying the error and improving the Python script, the page navigation bar appears on the second page as well.

navigation_bug


doomedraven:
interesting, thank you

PR #1333, opened 2023-01-26
add more options to the installer

VVelox:
ClamAV support can now be enabled via an option instead of having to call separately.

MongoDB(and by extension the web stuff) can now be disabled. Defaults to on.

LibreNMS support is now available for it. Various options for that. This will enabled snmpd.

On a related note, finalizing Sneck support in LibreNMS is on my todo list, but this does configure it. Sneck basically exists to get around the massive short comings of NRPE as well as provide something that integrates nicely with LibreNMS.

### DIFF
--- a/installer/cape2.sh
+++ b/installer/cape2.sh
@@ -239,7 +239,6 @@ function librenms_snmpd_config() {
 		echo "#extend mdadm /usr/bin/env PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/sbin:/usr/local/bin /etc/snmp/extends/mdadm"
 	fi
 	echo
-	echo 'agentaddress 127.0.0.1'
 	if [ ! -z "$snmp_agentaddress" ]; then
 		echo "agentaddress $snmp_agentaddress"
 	fi
--- a/installer/cape2.sh
+++ b/installer/cape2.sh
@@ -190,8 +190,8 @@ function librenms_sneck_config() {
 	else
 		echo '#ipmi_sensor|/usr/lib/nagios/plugins/check_ipmi_sensor --nosel'
 	fi
-	echo 'virtqemud_procs|/usr/lib/nagios/plugins/check_procs --ereg-argument-array "^/usr/sbin/virtqemud" 1:2'
-	echo 'cape_procs|/usr/lib/nagios/plugins/check_procs --ereg-argument-array "poetry.*bin/python cuckoo.py" 1:2'
+	echo 'virtqemud_procs|/usr/lib/nagios/plugins/check_procs --ereg-argument-array "^/usr/sbin/virtqemud" 1:1'
+	echo 'cape_procs|/usr/lib/nagios/plugins/check_procs --ereg-argument-array "poetry.*bin/python cuckoo.py" 1:1'
 	echo 'cape_processor_procs|/usr/lib/nagios/plugins/check_procs --ereg-argument-array "poetry.*bin/python process.py" 1:'
 	echo 'cape_rooter_procs|/usr/lib/nagios/plugins/check_procs --ereg-argument-array "poetry.*bin/python rooter.py" 1'
 	if [ "$clamav_enable" -ge 1 ]; then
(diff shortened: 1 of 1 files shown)


installer/cape2.sh
```diff
@@ -604,43 +741,44 @@ function install_yara() {
...
-    if grep -q ' avx ' /proc/cpuinfo; then
-        MONGO_VERSION="6.0"
-    else
-        echo "[-] Mongo >= 5 is not supported"
-        MONGO_VERSION="4.4"
-    fi
-
-    sudo curl -fsSL "https://www.mongodb.org/static/pgp/server-${MONGO_VERSION}.asc" | sudo gpg --dearmor -o /etc/apt/keyrings/mongo.gpg --yes
-    echo "deb [signed-by=/etc/apt/keyrings/mongo.gpg arch=amd64] https://repo.mongodb.org/apt/ubuntu $(lsb_release -cs)/mongodb-org/${MONGO_VERSION} multiverse" > /etc/apt/sources.list.d/mongodb.list
+	if ( "$mongo_enable" -ge 1 ); then
```
ethhart, line 744: Should be brackets:
`if [ "$mongo_enable" -ge 1 ]; then`

VVelox: Ohh! Derp! Thanks. Yeah, was just about to find that here as well as just getting ready to test this all via ansible. =^.^=

VVelox: added fix to a  PR in https://github.com/kevoreilly/CAPEv2/pull/1334

installer/cape2.sh
```diff
@@ -604,43 +741,44 @@ function install_yara() {
...
+		# Mongo >=5 requires CPU AVX instruction support https://www.mongodb.com/docs/manual/administration/production-notes/#x86_64
+		if grep -q ' avx ' /proc/cpuinfo; then
+			MONGO_VERSION="6.0"
+		else
+			echo "[-] Mongo >= 5 is not supported"
+			MONGO_VERSION="4.4"
+		fi
+
+		sudo curl -fsSL "https://www.mongodb.org/static/pgp/server-${MONGO_VERSION}.asc" | sudo gpg --dearmor -o /etc/apt/keyrings/mongo.gpg --yes
+		echo "deb [signed-by=/etc/apt/keyrings/mongo.gpg arch=amd64] https://repo.mongodb.org/apt/ubuntu $(lsb_release -cs)/mongodb-org/${MONGO_VERSION} multiverse" > /etc/apt/sources.list.d/mongodb.list
```
ethhart, line 755: I'd also recommend something more like
`echo "deb [ arch=amd64,arm64 ] https://repo.mongodb.org/apt/ubuntu focal/mongodb-org/6.0 multiverse" | sudo tee /etc/apt/sources.list.d/mongodb-org-${MONGO_VERSION}.list`
I would imagine you would get a permission denied for most systems. 
https://www.mongodb.com/docs/manual/tutorial/install-mongodb-on-ubuntu/

VVelox: Using sudo won't really do anything meaningful here given this all needs to run as root any ways. Really no good reason to call sudo from this script unless you need to run it as something other than root.

installer/cape2.sh
```diff
@@ -604,43 +741,44 @@ function install_yara() {
...
+			sudo apt install sysfsutils -y
+			echo "kernel/mm/transparent_hugepage/enabled = never" >> /etc/sysfs.conf
+			echo "kernel/mm/transparent_hugepage/defrag = never" >> /etc/sysfs.conf
+		fi
+
+		if [ -f /lib/systemd/system/mongod.service ]; then
+			systemctl stop mongod.service
+			systemctl disable mongod.service
+			rm /lib/systemd/system/mongod.service
+			rm /lib/systemd/system/mongod.service
```
ethhart, line 775: Duplicate of above?

installer/cape2.sh
ethhart: Yeah, the sudo wasn't the point with that command. My pipeline was complaining about not having permissions to run that line, but it was due to something else. This one is resolved. Sorry.

PR #1334, opened 2023-01-27
add items to ignore that are created by utils/community.py to .gitignore

VVelox:
Minor cleanup here so git ignores stuff pulled in via utils/community.py -waf


VVelox:
Also pushed a small fix for the install scripted as well where I used if ( ... ) instead of if [ ... ].

Thanks @ethhart .

ethhart:
Can we use globs for the .gitignore to keep it cleaner and automatically include future files?

VVelox:
> Can we use wildcards for the .gitignore to keep it cleaner and automatically include future files?

Very much agreed.

That said as it currently stands, not aware of any directory that that does not also have stuff included in it from the base.

If you know of any or spot any, definitely point it out. :)

ethhart:
yeah. Realized that once I posted. Eventually we should shove those in their own community directory and traverse or copy them to keep them separate.

doomedraven:
so what is the reason to have community in cape's ignore?

doomedraven:
i have fixed installer by hand

VVelox:
> so what is the reason to have community in cape's ignore?

IMHO it is nice as so if you do a `git status` you don't get this bombed with a bunch of items you don't care that about not being in git.

TheMythologist:
You can manually add it `.git/info/exclude`, but I don't think this should be part of `.gitignore`

VVelox:
> You can manually add it `.git/info/exclude`, but I don't think this should be part of `.gitignore`

VVelox:
> You can manually add it `.git/info/exclude`, but I don't think this should be part of `.gitignore`

Right, but there is no good reason not to have to manually do this though as it just makes in every manner to work on/manage as you have to keep closer track of what you've added or the like v. what was dumped into place via the community script.

That said may be possibly a better idea to auto generate that via community script.

TheMythologist:
Here are my main concerns:

1. This is the wrong usage of `.gitignore`
    > .gitignore file lists all of the files that are local to a project that Git should not push to GitHub

    What if someone is working on the community signatures and not on CAPE itself? If this script is automatically done in cape's installer, he/she would be pretty confused

2. Even if this isn't done automatically and instead the user has to explicitly allow this adding to `.git/info/exclude`, who will be checking the community repo to update this "automatic" updating?
3. What if, for whatever reason, there are files in community that overwrite CAPEv2's files? (this has happened before, and it probably won't be the last time)

The best solution would be to use a proper project structure where we don't have to rely on writing files to the same directory structure that is handled by git (e.g. symlinking / pointing modules/signatures to `/opt/capev2_signatures` and other directories instead), but at this point I think the project structure is way too messy to configure this. And there are other tasks that require higher priority.

VVelox:
1: The community repo and this repo are two different items. So this is not relevant.

2/3: You seem to be missing the point of automatic generation. It handles both of these if you integrate it with the update script. In fact this enables nice automated detection and avoidance for three.

That said closing as this should be automated.

PR #1353, opened 2023-02-06
Update BumbleBee.yar

r0ny123:


kevoreilly:
Not sure why this replaces an existing working pattern $antivm3 instead of adding a new patter $antivm4. Also there is no update to the monitor yara to bypass the anti-vm which is the reason these patterns are picked, so replacing antivm3 in the monitor yara would break detonation for those samples that depended on it.

r0ny123:
That was not working.

r0ny123:
@kevoreilly, fixed the payload part. Can confirm nothing breaks.

PR #1364, opened 2023-02-10
Add script that downloads YARA from CAPE repo, can be turned into a c…

cccs-kevin:
…ronjob if you wish

Using this, you don't have to update the whole repo


kevoreilly:
Looks good, although worth noting the analyzer yaras are sometimes dependent on recent monitor builds, if a new feature has been added like the recent automated export picking for example.

PR #1366, opened 2023-02-13
Improve handling of signature call data.

tbeadle:
- Unconditionally *don't* store the call details in the "signatures" key of the analysis collection. Instead, only store the pid (process ID) and cid (chunk ID) of the call. That information alone allows us to query the behavior.processes.calls object for the ObjectID of the chunk and the index of the specific call within that chunk that triggered the signature.
- When rendering the signatures, don't immediately render the call data. Instead, use an AJAX call when a signature is expanded to query for its rendered HTML. To maintain backwards compatibility with documents that already contain the embedded call details, return the rendered HTML using the same template as before. Otherwise, we'll query Mongo or ES for the signature's call details first.
- Note that this also changes the display of signatures' call details to show it in tabular form like on the "Behavioral Analysis" tab.

Screenshot_20230213_162824

### DIFF
--- a/conf/cuckoo.conf
+++ b/conf/cuckoo.conf
@@ -91,11 +91,6 @@ max_len = 196
 sanitize_len = 32
 sanitize_to_len = 24
 
-# Add apicall details to signature?
-# Enabling this can result in data loss if using MongoDB as it can generate a lot of data.
-# See https://github.com/kevoreilly/CAPEv2/issues/1340.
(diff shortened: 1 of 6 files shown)


doomedraven:
thank you for rewriting it, i will give it a shot tomorrow

kevoreilly:
❤️

r0ny123:
This is cool!

PR #1369, opened 2023-02-15
Add missing guest names for dropped files

enzok:


doomedraven:
we will need to add tests for CAPE.py, as is become a core of many parts

PR #1370, opened 2023-02-15
Fix msi extraction bug

enzok:


doomedraven:
thanks, we have to add some tests

PR #1373, opened 2023-02-15
Allow a machine to be marked as "reserved" so that it can only be used if specifically requested.

tbeadle:
It a machine has a config that includes "reserved = yes" in the <machinery>.conf file, then that machine will only be used for a task if that machine was selected specifically for that task by the user. This could be useful if you have, say, a new VM that you want to be able to try out but you don't want it in the normal pool.

Note that this includes a database migration that adds a uniqueness constraint to the 'name' and 'label' columns in the 'machines' table. This seems logical to me, although it did require some test modifications and I'm not sure if users have multiple machines with the same name/label. It seems like other things would break if that were the case.

An alternate possible solution to this is using tags. I opted not to do that because it would likely require the use of a special tag like "reserved" that behaves unlike any other tag, which seemed like it could cause confusion.


nbargnesi:
I've got a small fix to the DB schema in this PR: https://github.com/tbeadle/CAPEv2/pull/1

doomedraven:
Hello both, i like this idea, once i will feel better i will test that on my dev server(bug bulletproof :D)

doomedraven:
thank you, looks good to me, at the moment it goes to staging branch

PR #1392, opened 2023-02-23
Using pipes instead of commas since get_options cannot handle comma d…

cccs-kevin:
…elimiters per option

https://github.com/kevoreilly/CAPEv2/blob/9f252cc4817977055188a4590a106e5229243bfb/lib/cuckoo/common/utils.py#L853

### DIFF
--- a/.yara-ci.yml
+++ b/.yara-ci.yml
@@ -6,3 +6,4 @@ files:
 false_positives:
   ignore:
     - rule: "CobaltStrikeBeacon"
+    - rule: "Emotet"
--- a/analyzer/windows/modules/packages/dll.py
+++ b/analyzer/windows/modules/packages/dll.py
@@ -58,9 +58,9 @@ class Dll(Package):
         run_multiple_functions = False
 
         if function:
-            # If user has requested we use functions (by name or by ordinal number), separated by commas
-            if enable_multi and "," in function:
-                function = function.split(",")
+            # If user has requested we use functions (by name or by ordinal number), separated by pipes
+            if enable_multi and "|" in function:
+                function = function.split("|")
                 run_multiple_functions = True
 
             # If user has requested we use an ordinal range, separated by a hyphen or by ..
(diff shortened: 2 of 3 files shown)


doomedraven:
so that would be function=1|2|3?

cccs-kevin:
> so that would be function=1|2|3?

Correct!

kevoreilly:
But aren't colons used for this elsewhere, e.g. exclude-apis=func1:func2

doomedraven:
ah yes that true

doomedraven:
i have replaced pipe by colon to keep it consistent

kevoreilly:
Thanks 👍

cccs-kevin:
Thanks team!

kevoreilly:
I do wonder if some more brainstorming could be useful on the tricky subject of 'hidden' exports - looks like my recent feature for using yara to select an export will be undermined by the AL solution picking another export first.

cccs-kevin:
> I do wonder if some more brainstorming could be useful on the tricky subject of 'hidden' exports - looks like my recent feature for using yara to select an export will be undermined by the AL solution picking another export first.

This commit? https://github.com/kevoreilly/CAPEv2/commit/0b1a634709535ce004475806968382fe349ef780

kevoreilly:
Well that is one example of the technique. To understand its mechanism the relevanr capemon commit is: https://github.com/kevoreilly/capemon/commit/64d21309ad498819a678e640324c71c5feeba93b

This mechanism relies on the dll being submitted with the default export '#1' and then allowing an address set by monitor yara to override the return value from the LdrGetProcedureAddress hook.

kevoreilly:
@doomedraven also uses another solution at his $dayjob whereby multiple exports are executed in multi-threaded fashion by a custom compiled loader - this is more similar to the AL logic as it relies on brute force or other more generic tests. The ideal would surely be to allow for both mechanisms to work alongside each other harmoniously.

cccs-kevin:
Right right. The AL solution is what our reverse engineering team came up with to cover all Dlls that we receive for analysis. I agree, it would ideal to have a "best of both worlds" solution.

PR #1393, opened 2023-02-23
Update uwsgi cape.ini

MayerDaniel:
update the uwsgi .ini file to have parity with the config in the documentation: https://capev2.readthedocs.io/en/latest/usage/web.html#best-practices-for-production


doomedraven:
thanks, for production it better to put nginx in front of uwsgi, specially if is internet facing

doomedraven:
thanks

PR #1390, opened 2023-02-23
CAPE 2.4

doomedraven:

PR #1396, opened 2023-02-24
Changing the column type of Task.options to Text with no length limit

cccs-kevin:
When submitting multiple exports to be executed in a Dll (fixed in https://github.com/kevoreilly/CAPEv2/pull/1392 and https://github.com/kevoreilly/CAPEv2/pull/1390/commits/8c9e9182ec8c1ef8b97b2908e44a0027fb6d18d4), the "options" field can get larger than 1024, therefore convert to Text so that we can add these tasks to the database.

This is on par with how Cuckoo handles the options column https://github.com/cuckoosandbox/cuckoo/blob/master/cuckoo/core/database.py#L331

### DIFF
--- a/lib/cuckoo/core/database.py
+++ b/lib/cuckoo/core/database.py
@@ -114,7 +114,7 @@ if repconf.elasticsearchdb.enabled:
 
     es = elastic_handler
 
-SCHEMA_VERSION = "d6aa5d949b70"
+SCHEMA_VERSION = "a8441ab0fd0f"
 TASK_BANNED = "banned"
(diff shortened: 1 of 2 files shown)


doomedraven:
I will review once im back, i also need to drop the parent id field

doomedraven:
ok here we have a small problem, fixing it

```
02af0b0ec686 -> a8441ab0fd0f (head), 2_4_0_change_options_to_text
02af0b0ec686 -> d6aa5d949b70 (head), Add 'reserved' column for machines
```

PR #1409, opened 2023-03-02
FIXED bug .xlsx

AndreaF17:
This is the code working in my prod env prod


doomedraven:
that was already fixed in different and better way, thanks for headups

PR #1413, opened 2023-03-02
Added: CSRF Protection

AndreaF17:
I had some problems during the submit of the files using dns-name:

With this code you can get rid of the CSRF error if you set the allowed hostname in `CAPEv2/config/web.conf` as follows:
```
# hostname of the cape instance
hostname = https://example.com
```

I hope that helps

### DIFF
--- a/web/web/local_settings.py
+++ b/web/web/local_settings.py
@@ -4,7 +4,6 @@
 
 LOCAL_SETTINGS = True
 from .settings import *  # noqa: F403
-from lib.cuckoo.common.config import Config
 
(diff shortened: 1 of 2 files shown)


doomedraven:
Thanks this is useful, but should go into settings and not local, can you move it? Local is for overwrite/extend settings. Py

AndreaF17:
Yes, no worries. I'm going to do it!

AndreaF17:
That should be it!

doomedraven:
thank you

PR #1415, opened 2023-03-06
deps: remove unnecessary dependencies

TheMythologist:
These dependencies were already commented. Those that are completely not referenced at all (both in https://github.com/kevoreilly/CAPEv2 and https://github.com/CAPESandbox/community) are removed.

Other changes: verbose print statement in `vbadeobf.py` was removed.


doomedraven:
Thank you

PR #1431, opened 2023-03-10
File trimming - PE, OLE Doc

enzok:


lib/cuckoo/common/web_utils.py
```diff
@@ -1249,16 +1249,15 @@ def process_new_task_files(request, samples, details, opt_filename, unique):
...
                     trimmed_size = trim_sample(sample.chunks().__next__())
                     if trimmed_size:
                         size = trimmed_size
-                        data = sample.chunks(size).__next__()
-                else:
-                    # we need to rebuild original file
-                    data = first_chunk + sample.read()
+                # move to start of sample file
+                sample.seek(0)
+                data = sample.read(size)
```
tbeadle, line 1260: I think these 2 lines should go after the following 'if' block so that we don't read a huge file in to memory if we're just going to skip it anyway.

lib/cuckoo/common/trim_utils.py
```diff
@@ -0,0 +1,56 @@
...
+
+from lib.cuckoo.common.config import Config
+from lib.cuckoo.common.objects import File
+from lib.cuckoo.common.integrations.parse_pe import PortableExecutable
+from lib.cuckoo.common.path_utils import path_write_file
+
+web_cfg = Config("web")
+
+
+def trim_file(filename: bytes, options: str, doc: bool = False, chunk: bytes=False, return_size:bool=False) -> bool:
```
tbeadle, line 13: I don't see where the 'chunk' or 'return_size' parameters ever get passed in.

tbeadle (commented):
Thanks for working on this. The only concern I have is that the hashes that will be recorded for the sample are going to be that of the trimmed sample and not the original, in which case it's not going to line up with other things that reference that sample, such as VT. Do you have any thoughts around this?

lib/cuckoo/common/web_utils.py
```diff
@@ -1249,16 +1249,15 @@ def process_new_task_files(request, samples, details, opt_filename, unique):
         size = sample.size
         data = False
         if size > web_cfg.general.max_sample_size:
-            if not (web_cfg.general.allow_ignore_size and "bypass_size_check" in details["options"]):
+            if not (web_cfg.general.allow_ignore_size and "ignore_size_check" in details["options"]):
                 first_chunk = sample.chunks().__next__()
                 if web_cfg.general.enable_trim and HAVE_PEFILE and IsPEImage(first_chunk):
```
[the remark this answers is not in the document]
enzok: my original PR was handling OLE

lib/cuckoo/common/web_utils.py
enzok: Agreed.

lib/cuckoo/common/demux.py
```diff
@@ -217,9 +217,17 @@ def demux_sample(filename: bytes, package: str, options: str, use_sflock: bool =
     # TODO: Remove after checking all uses of demux_sample use bytes ~TheMythologist
     if isinstance(filename, str) and use_sflock:
         filename = filename.encode()
+
     # if a package was specified, then don't do anything special
     if package:
-        return [filename]
+        retlist = [filename]
+        if File(filename).get_size() > web_cfg.general.max_sample_size and not (
```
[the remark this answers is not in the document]
enzok: this is definitely easier

lib/cuckoo/common/trim_utils.py
enzok: looks like this is still work in progress - i have some updates that I will push later for review

tests/test_yara.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+import yara
+
+
+@pytest.mark.skip(reason="Not finished yet")
```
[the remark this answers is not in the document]
doomedraven: no it shouldn't i was about to remove it from here

doomedraven:
about hashes, i guess the best would be register original file as parent and then link the trimmed one so in that case we would keep relation between them

enzok:
Reworked this quite a bit. All trimming is done in demux.py since all submissions are passed through it anyway, and we are downloading the original file and linking its details if trim is enabled. Comments or additional improvements are welcome.

doomedraven:
last word is on @kevoreilly , btw removing `sample_parent_id` we lose track of initial binary and will have only trimmed, wouldn't it be good to set sample_parent_id?

kevoreilly:
Yes it would. Also I hope to add the doc trimming to the in-vm code.

enzok:
We still keep the parent. The only time trim parent will be lost is if the large doc is submitted in a zip that is demux'd. In this case the parents will be the zip(or other archive.)

PR #1448, opened 2023-03-20
bugfix: azure - don't set spot_restore_policy unless using Spot

tbartlett0:
Azure machinery is setting the spot_restore_policy setting even when az.spot_instances option is set to false. This causes Azure to return an error and the VM scale set to not be created.

Update az.py to set spot_restore_policy to None if not using spot instances.


doomedraven:
@cccs-kevin can you review please?

cccs-kevin (approved):
Thank you @tbartlett0 !

tbartlett0:
Cheers 👍

PR #1452, opened 2023-03-22
Fix guacamole version and poetry path for guac-web.service

redteampanda-ng:
Current version 1.4.0 is not available on the remote server anymore, throwing a 404 error and effectively not installing guacamole.
Additionally add a check for poetry path, since the guac-web.service will not work out of the box


doomedraven:
thanks

PR #1456, opened 2023-03-22
Update how sample hashes are associated with extracted configs and display them in the UI.

tbeadle:
If more than one artifact from a detonation has the same config extracted from it, associated the hashes for each of them with the config. This requires the _associated_config_hashes to be stored as a list of dicts instead of a single dict.

Rename associated_config_hashes and associated_analysis_hashes to each have a leading underscore. This makes it easier to identify what is a malware family (no leading underscore) from something that is more like meta data. The alternative would've been to include the hashes inside the family's config dict, but I think it's better to keep that to only the actual config pieces that were extracted.

UI: For each extracted config, display the sha256 of each artifact it was extracted from. Clicking on it will expand to show the other hashes for the artifact.

Here's what the UI looks like with the hashes collapsed:
collapsed
Note that there could be more than one sha256 listed.

Clicking on the sha256 expands it out like this:
expanded

### DIFF
--- a/lib/cuckoo/common/objects.py
+++ b/lib/cuckoo/common/objects.py
@@ -548,7 +548,7 @@ class File:
         filetype = self.get_content_type()
         return any([ftype in filetype for ftype in type_list])
 
-    def get_all_hashe(self):
+    def get_all_hashes(self):
         return {
             "crc32": self.get_crc32(),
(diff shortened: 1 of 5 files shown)


doomedraven:
thanks, i will merge that to staging at the moment

PR #1459, opened 2023-03-24
Minor fixes

federicofantini:
- CSRF_TRUSTED_ORIGINS django SystemCheckError
- suricata cannot write log files
- fixed wrong f-string in TLS master secrets file


kevoreilly:
Thanks for this!

PR #1460, opened 2023-03-24
If free=yes, a pid in pids returned from analysis package can be None

cccs-kevin:
This was causing the analyzer to crash and not perform analysis cleanup.


kevoreilly:
Thank you

PR #1465, opened 2023-04-03
Hancitor update

cccs-j:
Implemented code from OALabs (source: https://github.com/OALabs/Lab-Notes/blob/main/Hancitor/hancitor.ipynb) to use regex to find the encryption key instead of hard coded offset from the  .data section.

A few samples have the key in the .text section causing the extractor to fail: 137ae44ea6f625a3dae762a668b68c96b10f953e5b407d3f5600da096a014f46 1a495667c6a44520eaa6f284a254973e0ea0cf222e37ab4bb06a9b5cf930d89c

### DIFF
--- a/modules/processing/parsers/CAPE/Hancitor.py
+++ b/modules/processing/parsers/CAPE/Hancitor.py
@@ -3,6 +3,8 @@
 """
 import hashlib
 import logging
+import re
+import struct
 
 import pefile
 from Cryptodome.Cipher import ARC4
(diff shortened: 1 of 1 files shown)


kevoreilly:
This looks great - thanks a lot! Will test & merge asap.

doomedraven:
isn't chanitor/hancitor is dead family?

kevoreilly:
I am working on this PR but unfortunately the 2 samples mentioned do not detonate and are also missing detection so I'm doing a more thorough survey before merging to make sure detection/unpacking are also working. So far the most recent sample I can find is from about a year ago in https://www.malware-traffic-analysis.net/2022/03/21/index2.html

kevoreilly:
Unfortunately I have discovered that these changes break config extraction for previously working samples, for example `abc91dfef90476ff6e3406b62b085ef1e84585aea3e9d80dbe3aa21597ff5995` which is the sample from https://www.malware-traffic-analysis.net/2022/02/01/index.html

Also as noted above, the sample hashes supplied with this PR do not detonate, and appear more likely to be unpacked/memory dumps than original samples. I would ideally like to find the original packed samples, or other working examples of samples for which the old parser failed, in order to test the full chain of config extraction.

cccs-js:
> 

not very active, but I wouldn't call this completely dead either:
https://bazaar.abuse.ch/browse.php?search=tag%3Ahancitor
last dll on there is 2022-03-21

kevoreilly:
Right I've integrated this new key extraction method into the parser, so thank you for contributing this code. I found that the regular expression search failed to find the key in the unpacked payload from `abc91dfef90476ff6e3406b62b085ef1e84585aea3e9d80dbe3aa21597ff5995`. I also found that for the two retrohunt samples there was nothing in the hard-coded location in `.data`. Therefore to keep things working that were previously, but also handle the retrohunt samples, I added a check for empty key data from the hardcoded location, which invokes the new key search.

This now works for all previous samples and the mentioned retrohunt samples. I made a mess of attempting to update the PR so decided to keep that mess from the main repo and just push the changes directly in https://github.com/kevoreilly/CAPEv2/commit/fbafc08deba346405388eed98d24b6f559b70eaa

Thanks again for contributing to cape.

PR #1478, opened 2023-04-10
Add "processed_on" field to analysis info.

tbeadle:
This field will help determine if a task has been reprocessed, allowing external systems to know if the document may have been updated since the last time it was polled.


kevoreilly:
No good?

tbeadle:
It could certainly go in, but it doesn't actually suit our needs at the moment. I had to approach it a different way.

PR #1480, opened 2023-04-11
Fix Service DLL Package

kevoreilly:


oalieno:
I notices that I forgot to remove the log in my code snippet. For example, `log.info("wow64_redirection")`. It is better to remove them before merging it.

kevoreilly:
Yes I must have made a mess of git yet again - in order to simply add just the wanted changes, I resorted to just pushing them in a single commit: https://github.com/kevoreilly/CAPEv2/commit/6437b84965ee0ca5d9918978a78b332524290105

Thanks for your contribution and sorry about the delay getting it merged,

PR #1484, opened 2023-04-14
proxy/vpn selection feature

federicofantini:
Hi, in my use case it is convenient for me to be able to select which route to submit to the analysis and therefore I have added this feature. I hope it can be useful for others too :)

I also have a question: which socks5 transparent proxy do you use? I tried several socks5 transparent proxies (redsocks, redsocks2, gost, danted...) but your `iptables` double REDIRECT target for UDP and TCP traffic only works with: https://github.com/n1nj4sec/pr0cks.

NOTES: I also modified the `utils/submit.py` file because I couldn't submit the analysis correctly, if I touched something wrong, tell me and I'll change it back :+1:

### DIFF
--- a/lib/cuckoo/core/rooter.py
+++ b/lib/cuckoo/core/rooter.py
@@ -12,9 +12,8 @@ cfg = Config()
 router_cfg = Config("routing")
 log = logging.getLogger(__name__)
 
-vpns = {}
-socks5s = {}
-
+vpns = dict()
+socks5s = dict()
 
 def _load_socks5_operational():
 
@@ -42,6 +41,14 @@ def _load_socks5_operational():
                 continue
 
             socks5s[name] = socks5.to_dict()
+
+            # decode utf-8 socks5man database data
+            for k, v in socks5s[name].items():
+                if isinstance(v, (bytes, bytearray)):
+                    try:
+                        socks5s[name][k] = v.decode()
+                    except (UnicodeDecodeError, AttributeError):
+                        pass
     except Socks5manDatabaseError as e:
         print(e, "you migth have an outdated database at $HOME/.socks5man")
 
--- a/web/submission/views.py
+++ b/web/submission/views.py
@@ -34,7 +34,7 @@ from lib.cuckoo.common.web_utils import (
     process_new_task_files,
 )
 from lib.cuckoo.core.database import Database
-from lib.cuckoo.core.rooter import _load_socks5_operational, vpns
+from lib.cuckoo.core.rooter import _load_socks5_operational
 
 # this required for hash searches
 cfg = Config("cuckoo")
@@ -548,15 +548,11 @@ def index(request, task_id=None, resubmit_hash=None):
         vpn_random = ""
 
         if routing.socks5.random_socks5 and socks5s:
-            socks5s_random = random.choice(socks5s.values()).get("name", False)
+            socks5s_random = socks5s[random.choice(list(socks5s.keys()))]
 
-        if routing.vpn.random_vpn:
-            vpn = list(vpns.values())
-            if vpn:
-                vpn_random = random.choice(vpn).get("name", False)
-
-        if socks5s:
-            socks5s_random = random.choice(list(socks5s.values())).get("name", False)
+        from lib.cuckoo.core.rooter import vpns
+        if routing.vpn.random_vpn and vpns:
+            vpn_random = vpns[random.choice(list(vpns.keys()))]
 
         random_route = False
         if vpn_random and socks5s_random:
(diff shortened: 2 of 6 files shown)


web/submission/views.py
```diff
@@ -548,15 +548,11 @@ def index(request, task_id=None, resubmit_hash=None):
...
+            socks5s_random = socks5s[random.choice(list(socks5s.keys()))]
 
-        if routing.vpn.random_vpn:
-            vpn = list(vpns.values())
-            if vpn:
-                vpn_random = random.choice(vpn).get("name", False)
-
-        if socks5s:
-            socks5s_random = random.choice(list(socks5s.values())).get("name", False)
+        from lib.cuckoo.core.rooter import vpns
```
doomedraven, line 553: is not good to import this over and over one page load

doomedraven:
hey redsocks2+dante works fine for me with current setup

web/submission/views.py
federicofantini: ops, fixed it!

federicofantini:
So you use redsocks2 as local socks5 transparent proxy and dante as remote socks5 proxy right? You listen with redsocks2 on the same interface as libvirt and with the double iptables REDIRECT you forward both UDP and TCP traffic.
Redsocks2 always gave me "dropping client" errors for all UDP traffic but works fine with TCP traffic, may I see your UDP configuration? :)

doomedraven:
>So you use redsocks2 as local socks5 transparent proxy and dante as remote socks5 proxy right?

yes

> You listen with redsocks2 on the same interface as libvirt 

no, they are on remote servers, but still it all kinda transparent as it works as tor socks5 socks, so it shouldn't really matter where they are

```
redudp {
    // `local_ip' should not be 0.0.0.0 as it's also used for outgoing
    // packets that are sent as replies - and it should be fixed
    // if we want NAT to work properly.
    local_ip = X;
    local_port = X;

    // `ip' and `port' of socks5 proxy server.
    ip = X;
    port = X;
    //login = username;
    //password = pazzw0rd;
    type = socks5;
    dest_ip = X;
    dest_port = 53;
    udp_timeout = 30;
    udp_timeout_stream = 180;
}
```

federicofantini:
> no, they are on remote servers, but still it all kinda transparent as it works as tor socks5 socks, so it shouldn't really matter where they are

@doomedraven _I'll try to explain myself better, if I'm doing something wrong please tell me._

By "local" I'm referring to the "local_ip" and "local_port" clause which must necessarily listen on the IP of the virtual network created by libvirt, for example `192.168.100.1/24` port `12345`. Necessarily because the REDIRECT target of iptables changes the destination IP to the primary address of the incoming interface -> libvirt iface.
So redsocks2 manages to act as a local transparent socks5 proxy sending all traffic to the remote server specified below.

A small schema of the networking path:
1. **sandbox vm** (`192.168.100.111/24`) try to connet to `8.8.8.8:53`
2. **iptables REDIRECT** changes dest to `192.168.100.1/24` and port `12345`
3. **local transparent socks5 proxy** (`192.168.100.1/24` and port `12345`) makes a UDP associate request over TCP to the remote socks5 proxy server (IP:PORT), the remote server choose the port and finally the real UDP datagram is sent
4. **remote socks5 proxy** (IP:PORT) has received the real UDP datagram and sends it to the real destination `8.8.8.8:53`

lib/cuckoo/core/rooter.py
```diff
@@ -12,9 +12,8 @@
 router_cfg = Config("routing")
 log = logging.getLogger(__name__)
 
-vpns = {}
-socks5s = {}
-
+vpns = dict()
+socks5s = dict()
```
doomedraven, line 16: FYI for future if is already declared, don't change that as ruff and other will ask to restore from dict() to {}

lib/cuckoo/core/rooter.py
```diff
@@ -42,6 +41,14 @@ def _load_socks5_operational():
                 continue
 
             socks5s[name] = socks5.to_dict()
+
+            # decode utf-8 socks5man database data
+            for k, v in socks5s[name].items():
+                if isinstance(v, (bytes, bytearray)):
+                    try:
```
doomedraven, line 48: `with suppress(UnicodeDecodeError, AttributeError)`

doomedraven:
next time please enable option to allow owners edit your code, i can't push fixes here 

```
# Copyright (C) 2010-2013 Claudio Guarnieri.
# Copyright (C) 2014-2016 Cuckoo Foundation.
# This file is part of Cuckoo Sandbox - http://www.cuckoosandbox.org
# See the file 'docs/LICENSE' for copying permission.

import logging
from contextlib import suppress

from lib.cuckoo.common.config import Config
from lib.cuckoo.common.socket_utils import send_socket_command

cfg = Config()
router_cfg = Config("routing")
log = logging.getLogger(__name__)

vpns = {}
socks5s = {}

def _load_socks5_operational():

    socks5s = {}

    if not router_cfg.socks5.enabled:
        return socks5s

    try:
        from socks5man.exceptions import Socks5manDatabaseError
        from socks5man.manager import Manager
    except (ImportError, OSError):
        return socks5s
    except Exception as e:
        log.error(e)
        return socks5s

    try:
        for socks5 in Manager().list_socks5(operational=True):
            if not hasattr(socks5, "description"):
                continue

            name = socks5.description
            if not name:
                continue

            socks5s[name] = socks5.to_dict()

            # decode utf-8 socks5man database data
            for k, v in socks5s[name].items():
                if isinstance(v, (bytes, bytearray)):
                    with suppress(UnicodeDecodeError, AttributeError):
                        socks5s[name][k] = v.decode()

    except Socks5manDatabaseError as e:
        print(e, "you migth have an outdated database at $HOME/.socks5man")

    return socks5s


def rooter(command, *args, **kwargs):
    ret = send_socket_command(cfg.cuckoo.rooter, command, *args, **kwargs)
    if ret and ret.get("exception"):
        log.warning("Rooter returned error: %s", ret["exception"])
    return ret

```

doomedraven:
yes local should be in the libvirt network, so if your virbr1 is 192.168.100.0/24 then listen on .1 and port X

federicofantini:
Exactly, I have the same setup but redsocks2 for UDP traffic still doesn't work.
I keep investigating...

federicofantini:
I also tried more exotic/complicated setups with TPROXY but that didn't work either.
Only double redirect (TCP and UDP) + [pr0cks](https://github.com/n1nj4sec/pr0cks) works for me.

PR #1488, opened 2023-04-18
Loosen vidar yara condition

r0ny123:


doomedraven:
current yara gives a lot of FP, if we lose it more, that will be already a nightmare

doomedraven:
problem was on my side, just tested no fps

kevoreilly:
Excellent - thank you r0ny!

PR #1507, opened 2023-05-05
Vba2graph Fixed VBA2GRAPH integration

RoemIko:
After some digging and debugging found out that the vba2graph.py integration was deleted, It used to be in `lib/cuckoo/common/office/vba2graph.py` and was deleted in a commit, could not find the exact commit however.

I have tested it with the following file: `sha256:5b81f8f1208d2dfccb4dd6946102b61ad8f220c7b1c0a80f7be3ca23e6e59b3e`


doomedraven:
hm nice spot, thank you

doomedraven:
i have merged this to staging branch

PR #1520, opened 2023-05-15
Web: Fix Download All function

heck-gd:
"Download all as zip" is broken on various sub-pages. The `send_filename` needs to be determined differently if `paths` is a list.


doomedraven:
thanks

PR #1529, opened 2023-05-19
Fix Additional Filter in Behavioral Analysis

heck-gd:
This MR:
* Fixes using the Additional Filter in process tabs (it only worked in the first tab due to non-unique IDs).
* Fixes a bug where tid was casted to `int` but `call["thread_id"]` is always `str`.
* Removes requirement to specify both, caller and tid, allowing you to filter for just one of the two.


doomedraven:
Thank you, merged to staging branch for 1w

PR #1530, opened 2023-05-19
Rename doomedraven config files to cape in install scripts and prevent writing duplicate config lines

seanthegeek:


doomedraven:
You dont want to be backdoored by dommed and prefer cape? 🤣

seanthegeek:
@doomedraven  haha. Just making it clearer to new users what wrote the config files. Also, I just pushed another commit with more installer improvements.

doomedraven:
Thanks for updating it

PR #1537, opened 2023-05-23
Volatility configuration issue fix

sbjzn:
ERROR: Generic error executing volatility
  File "/opt/CAPEv2/utils/../modules/processing/memory.py", line 339, in run
    vol = VolatilityManager(self.memory_path)
  File "/opt/CAPEv2/utils/../modules/processing/memory.py", line 196, in __init__
    if isinstance(self.options.mask.pid_generic, int):
AttributeError: 'VolatilityManager' object has no attribute 'options'

It seems like the config for volatility is never fetched, adding the following lines resolves the issue.


doomedraven:
thank you

PR #1538, opened 2023-05-23
Add --enable-dotnet to yara build

seanthegeek:


doomedraven:
is not enabled bcz is enabled by default, please read docs properly
 "dotnet" module is enabled by default. It can be disabled with --disable-dotnet. https://github.com/VirusTotal/yara/blob/master/configure.ac#L215

PR #1540, opened 2023-05-25
Prevent exceptions in guac-web app

seanthegeek:
After authentication is enabled in the main Django app, the `guac-web` app will raise exceptions when trying to load a Guacamole console page. This PR updates `guac_settings.py` to avoid these issues.

Exceptions addressed.

> Model class django.contrib.sessions.models.Session doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS

> django.core.exceptions.ImproperlyConfigured: settings.DATABASES is improperly configured. Please supply the ENGINE value. Check settings documentation.

### DIFF
--- a/web/web/guac_settings.py
+++ b/web/web/guac_settings.py
@@ -77,7 +77,7 @@ TEMPLATES = [
 ]
 
 # Database settings. We don't need it.
-DATABASES = {"default": {"ENGINE": "django.db.backends.sqlite3", "NAME": "siteauth.sqlite"}
+DATABASES = {"default": {"ENGINE": "django.db.backends.sqlite3", "NAME": "siteauth.sqlite"}}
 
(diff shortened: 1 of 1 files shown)


doomedraven:
@enzok any thoughs?

seanthegeek:
To add more context, these exceptions occur after setting `enabled = yes` in the `web_auth` section of `conf/web.conf` ,  and then running 

```text
poetry run python3 manage.py migrate
poetry run python manage.py createsuperuser
```

The exceptions persist after the database migration, even if `web_auth` is disabled in the configuration file later.

This PR just adds two missing settings to `web/web/guac_settings.py` that are the same as `web/web/settings.py`, which don't cause any issues with either Django application, even when web authentication isn't enabled.

doomedraven:
thanks

PR #1546, opened 2023-05-26
Update objects.py

doomedraven:


lib/cuckoo/common/objects.py
```diff
@@ -361,6 +361,8 @@ def get_type(self):
         """
         if self.file_type:
             return self.file_type
+        else:
+            self.file_type = self.get_content_type()
```
tbeadle, line 365: technically, it's not necessary to put this in an 'else' block.

doomedraven: i have really no idea why Kevin changed that so i will leave this PR to be handle by him

PR #1549, opened 2023-05-27
url analysis package

doomedraven:


doomedraven:
@seanthegeek

seanthegeek:
That's exactly what I wanted! Thanks! I figured I was overthinking the code changes that needed to be made for that. I ned to learn more about the code structure.

Two things

Maybe make the default value `edge` instead of `chrome`? Although Chrome is more popular, Edge is included as a part of Windows 10 (and also has a release that supports Windows 7).

Why change the `web.conf.default` file and not also change the `web.conf` file for new clones/installs? users should be using the `custom/conf` directory to store their customized configuration, and if they changed the original config file, they could always stash their changes before pulling.

I've noticed a few other `.conf` and `.conf.default` files that are out of sync (I only checked the ones I use. There might be more.)

diff conf/cuckoo.conf conf/cuckoo.conf.default 

```diff
150a151,153
> # If you use PostgreSQL: SSL mode
> # https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-SSLMODE-STATEMENTS
> psql_ssl_mode = disable
```
diff conf/processing.conf conf/processing.conf.default 

```diff
116,117d115
< # Extract dotnet strings
< dotnet = no
```

diff conf/web.conf conf/web.conf.default 

```diff
40c40
< existent_tasks = yes
---
> existent_tasks = no
170a171,172
> # Server that exposes the VNC ports (e.g., your KVM host)
> vnc_host = localhost
```

doomedraven:
1. Ok
2. You are wrong. .Default is base config that always will be modded by new fields. The x.conf is modified by user. Custom/conf is backward compatibility. As it was confusing so many user, we become with conf.default so on gitpul it not generate headache
They are not out of sync. All base or how you call them missed are loaded from default config. That is your job to mod whatever you want in .conf to not suffer on sandbox updates

seanthegeek:
Ah, ok.

PR #1550, opened 2023-05-28
Remove the protocol from the hostname value in web.conf

seanthegeek:
`CSRF_TRUSTED_ORIGINS` in `web/web/settings.py` expects the `hostname` value in `conf/web.conf` to be the hostname, not the full URL.

https://github.com/kevoreilly/CAPEv2/blob/master/web/web/settings.py#L40

### DIFF
--- a/conf/web.conf.default
+++ b/conf/web.conf.default
@@ -40,8 +40,8 @@ anon_viewable = no
 existent_tasks = no
 top_detections = yes
 # hostname of the cape instance
-hostname = https://127.0.0.1/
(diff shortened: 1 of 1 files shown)


doomedraven:
that is user task, we doing only mods in web.conf.default which is modified

seanthegeek:
This is a PR to modify `web.conf.default`, which has an incorrect value.

doomedraven:
thanks, missed somehow default when was checking the filename

doomedraven:
i guess is time to review CENTS, as this value was introduced for cents

PR #1569, opened 2023-06-01
Update NGINX instructions

seanthegeek:
- Add `www-data` to `cape` group so NGINX can read recording files created by `guacd`
- Add warning about the lack of authentication on the CAPE `guac-web` service and recording files
-  Add optional NGINX HTTP basic authentication instructions
- Set `guacd.service` group back to cape
  - Revert change from https://github.com/kevoreilly/CAPEv2/commit/d722d7367ebed8ce72546fb1ffbff53d539ea8f0 because `guacd` will write files owned by the same group as the user either way

I understand that even without authentication, the risk of unauthorized access to Guacamole console and records are low, because the CAPE session ID and the Guacamole session GUID must be known to build the URL. That said, the ideal situation would be to build the functionality of the `guac-web` Django application into the main Django application, so the same authentication mechanism could be used for everything, and the recording files could be served behind that instead of being served directly by NGINX as static files.

I don't know enough about Django right now to do that myself but thought I should call out the (small) risk, and hope that maybe someone will take up the task of combining the web apps before I can, because I don't think I'll be able to get to it anytime soon.

### DIFF
--- a/docs/book/src/usage/interactive_desktop.rst
+++ b/docs/book/src/usage/interactive_desktop.rst
@@ -29,7 +29,7 @@ Enable and configure ``guacamole`` in ``conf/web.conf`` and restart ``cape-web.s
 
 In case you using are ``NGINX``, you need to configure it to be able to use interactive mode.  Here's an example config.
 
-Replace `www.capesandbox.com` with your own hostname.
+Replace ``www.capesandbox.com`` with your own hostname.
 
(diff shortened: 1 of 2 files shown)


doomedraven:
to merge it I need @enzok approve

enzok:
It's fine. Please see https://github.com/kevoreilly/CAPEv2/pull/1573 as well.

PR #1575, opened 2023-06-02
Update guac install and recordings config

enzok:
Latest guacd now adds group permissions to session recordings, so no need for fstab workaround.


seanthegeek:
`guacamole_version` also needs bumped to `1.5.2` because it is the only available version now.

PR #1590, opened 2023-06-08
dotnet

doomedraven:


lib/cuckoo/common/objects.py
```diff
@@ -375,15 +375,18 @@ def get_type(self):
                             is_dll = self.pe.is_dll()
                             is_x64 = self.pe.FILE_HEADER.Machine == IMAGE_FILE_MACHINE_AMD64
                             gui_type = "console" if self.pe.OPTIONAL_HEADER.Subsystem == 3 else "GUI"
+                            dotnet_string = (
+                                "Mono/.Net assembly" if self.pe.DIRECTORY_ENTRY_COM_DESCRIPTOR.VirtualAddress == 0 else ""
```
tbeadle, line 379: I just got the following traceback during a detonation:
```
   File "/opt/CAPEv2/lib/cuckoo/common/objects.py", line 379, in get_type
     "Mono/.Net assembly" if self.pe.DIRECTORY_ENTRY_COM_DESCRIPTOR.VirtualAddress == 0 else ""
AttributeError: 'PE' object has no attribute 'DIRECTORY_ENTRY_COM_DESCRIPTOR'
```

PR #1594, opened 2023-06-08
Updated docs to reflect pyw preference and some troubleshooting with poetry upgrade

RazviOverflow:
I think it is important to warn the used that the **standard, preferred** agent.py version is the cmd-windows-less *.pyw one.
Added some indications that may come handy when upgrading with poetry.


doomedraven:
thanks

PR #1598, opened 2023-06-09
Catch AttributeError when attempting to get DIRECTORY_ENTRY_COM_DESCIPTOR on non-CLR files

tbeadle:


doomedraven:
i prefer this solution, but thank you https://github.com/kevoreilly/CAPEv2/commit/a5db416a14bcbb64041a7193c2e5a0518b876eb4

PR #1601, opened 2023-06-12
Keep IPv6 enabled on the loopback interface

seanthegeek:
This change fixes compatibility with xrdp on the host system. xrdp's session manager, `xrdp-sesman` exclusively binds to a `tcp6` socket on the loopback interface, so disabling IPv6 on the loopback interface prevents xrdp from connecting to its session manager.

This change does not break per-analysis routing or cause leaks because it only applies to the loopback interface of the host system.


doomedraven:
hey im fine to merge this, but just as for context, what uses xrdp-sesman?

PR #1609, opened 2023-06-14
Add Chromium package

seanthegeek:
- Update `Package.enum_paths()` to support arbitrary environment variables as the path `basename`
- Add `chromium` package for Chromium browsers installed using the mini installer
  - See https://vikyd.github.io/download-chromium-history-version/ for release history and official download links
- Add `--disable-features=RendererCodeIntegrity` as a command line argument to `chrome.exe` to bypass the "Aw snap! `STATUS_INVALID_IMAGE_HASH` error
- Add chromium as a fallback browser for the chrome package, and chrome as the fallback for chromium
  -  Useful if a user accidently selects a package for a browser that is not installed


seanthegeek:
Apparently the cached venv for the workflow does not include pytest.

seanthegeek:
This PR is ready to be merged. However, with or without this PR, Chrome and Chromium are crashing when `capemon` is used. See issue #1636

kevoreilly:
Monitor issue should be fixed now - thanks for your contribution.

PR #1610, opened 2023-06-15
Use sudo to invoke tcpdump, if possible.

tbeadle:
Rather than having to worry about the capabilities set on tcpdump and maintaining it whenever tcpdump gets upgraded, just use sudo to invoke it. This retains backwards compatibility, so it will fall back to check if tcpdump is suid if it can't be run via sudo.
Update the installer script to add a file to /etc/sudoers.d to configure it.

### DIFF
--- a/installer/cape2.sh
+++ b/installer/cape2.sh
@@ -924,10 +924,8 @@ function dependencies() {
         useradd --system -g ${USER} -d /home/${USER}/ -m ${USER} --shell /bin/bash
     fi
 
-    groupadd pcap
-    usermod -a -G pcap ${USER}
-    chgrp pcap ${TCPDUMP_PATH}
-    setcap cap_net_raw,cap_net_admin=eip ${TCPDUMP_PATH}
+    echo "${USER} ALL=NOPASSWD: ${TCPDUMP_PATH}" > /etc/sudoers.d/tcpdump
+    chmod 440 /etc/sudoers.d/tcpdump
 
     usermod -a -G systemd-journal ${USER}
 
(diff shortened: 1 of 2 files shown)


heck-gd:
@tbeadle This broke our deployment. You shouldn't assume that `sudo` is installed (we're running CAPE containerized and there it's not available). Can you please follow up with a fix handling that case?

```
  File "/opt/CAPEv2/cuckoo.py", line 138, in <module>
    cuckoo_init(quiet=args.quiet, debug=args.debug, artwork=args.artwork, test=args.test)
  File "/opt/CAPEv2/cuckoo.py", line 86, in cuckoo_init
    check_tcpdump_permissions()
  File "/opt/CAPEv2/lib/cuckoo/core/startup.py", line 539, in check_tcpdump_permissions
    subprocess.check_call(["/usr/bin/sudo", "--list", "--non-interactive", tcpdump])
  File "/usr/local/lib/python3.10/subprocess.py", line 364, in check_call
    retcode = call(*popenargs, **kwargs)
  File "/usr/local/lib/python3.10/subprocess.py", line 345, in call
    with Popen(*popenargs, **kwargs) as p:
  File "/usr/local/lib/python3.10/subprocess.py", line 971, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/local/lib/python3.10/subprocess.py", line 1863, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/usr/bin/sudo'
```

PR #1615, opened 2023-06-19
Update/recentfiles

cccs-kevin:
Migrating the `recentfiles` auxiliary module from Cuckoo


doomedraven:
Thanks, we dont need to modify not default configs

PR #1620, opened 2023-06-21
interactive esxi

doomedraven:


doomedraven:
also @enzok what do you think about this update?

doomedraven:
i won't merge this, as libvirt allows to fetch all the data, but as i don't have ESXI access/setup, someone who wants to use it will need to extend it properly

Issue #1618, opened 2023-06-20
Memory amount reported by OS is not matching libvirt/KVM configuration


seanthegeek:
I think I've finally figured out the root cause of my VMs crashing - and why it didn't happen consistently when I changed different things to troubleshoot the issue - the guest OS does not see enough RAM.

The host has 64 GB of RAM.

Windows 10 64 bit only sees 2 GB of RAM on a VM configured with a Q35 chipset and 4 GB of RAM.

Windows 7 64 bit only sees 3 GB of RAM on a VM configured with an i440FX chipset and current and maximum amount of RAM set to 4 GB.

This happens with or without the virtio balloon driver installed (I'd prefer it to not have any virtio drivers installed, to make it harder to identify the system as a VM.

I realize this isn't a CAPE issue, but I'm hoping someone here has some idea of what's going wrong, because even after lots of Googling, and updating the BIOS of the host system, I still haven't figured this out.

FWIW, I also posted on the KVM Subreddit.  

Any idea what's going wrong?

doomedraven:
that is always bios, do grep on your bios file + watch the command like of qemu when vm start https://www.reddit.com/r/qemu_kvm/comments/phsrw8/cant_use_more_than_3gb_ram_on_x64_win10/

doomedraven:
im don't really like having not cape related problems in cape repo. https://github.com/doomedraven/Tools/blob/master/Virtualization/kvm-qemu.sh#L983-L987

seanthegeek:
@doomedraven No incidences from that grep command. I'm using the `bios.bin` built by the `seabios` command of the `kvm-qemu.sh` included with CAPE .

Now that you've pointed me in the direction of the BIOS though, I just tried using the default `bios.bin` and the VMs now show all 4 GB of RAM., so this does seem to be the opposite of what the help you quoted says .

By changing the smbios libvirt settings and adding the `qemu:commandline` element, like you mention in your [blog post](https://www.doomedraven.com/2016/05/kvm.html#2-add-cpuid-bit-vm-detect-patching-and-acpi-tables) I can pass just as many of the phfish VM checks, So what is the point of the custom BIOS build?

doomedraven:
1 software evolutivo maybe? They dropped al the antivm patterns

2 the point is to defeat all antivms, obviosly we wont check each release
of seabios

3 as my reddit says, in my case then it was the problem that the latest
easnt working, but previous worked perfectly

Welcome to software  updates hell 🤣

[...]

PR #1634, opened 2023-06-27
Update kvm-qemu.sh to address #1618

seanthegeek:
Exclude `paravirt.c` from patching to work around the issue described in #1618. I'm not sure if there is a better fix, but this will at least prevent VMs from crashing due to insufficient memory.


doomedraven:
thanks, the proper fix is to use the same MAGIC, but this works, thanks

PR #1643, opened 2023-07-04
Added CITATION.cff file to make it easier to cite CAPE

RazviOverflow:
GitHub automatically parses the file and transforms it in both APA and BibTeX formats:
image


doomedraven:
thank you

PR #1649, opened 2023-07-05
Yara Detections

para0x0dise:
# Main Issue: CAPE Check
CAPE performs a check by searching for an identifier called `cape_type` within the meta section of the **YARA** rule. If this variable contains specific words such as `payload`, `loader`, or `config`, CAPE adds the `"Detections"` section, which includes the name of the malware family.

image12

What if we want to download additional rules from multiple sources, it doesn't make any sense to add this identifier to all of them. To overcome this limitation, I modified this condition in the code to include the "Detections" section whenever a **YARA** hit is found.

image14

### DIFF
--- a/conf/processing.conf
+++ b/conf/processing.conf
@@ -208,6 +208,9 @@ max_file_size = 90
 userdb_signature = no
 # https://capev2.readthedocs.io/en/latest/usage/patterns_replacement.html
 replace_patterns = no
+# If enabled, the detections section will depend on YARA hits only if the rule contains an identifier named "cape_type" and contains "config", "payload", or "loader"
+# otherwise, it will be triggered whatever YARA hits are found
+yara_detections = no
 
 # Deduplicate screenshots
 # You need to install dependency ImageHash>=4.2.1
--- a/conf/processing.conf
+++ b/conf/processing.conf
@@ -208,9 +208,6 @@ max_file_size = 90
 userdb_signature = no
 # https://capev2.readthedocs.io/en/latest/usage/patterns_replacement.html
 replace_patterns = no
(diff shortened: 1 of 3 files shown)


doomedraven:
that is the beauty of the opensource, you can change whatever you want. I don't want to see all crap yara rules names in detection. but if you can extend conf/processing.conf.default [CAPE] to be able to on/off this feature. then i will merge it. if you don't know how, i will do that for you, just let me know

para0x0dise:
@doomedraven 
I extended "conf/processing.conf" to control the source of YARA rules, but to be noted, detections section depends on `data/yara/CAPE` which contains all rules related to malware families, so I think it will be not noisy.

para0x0dise:
@doomedraven 
I think we are good in this PR, may you review and merge it?

doomedraven:
we have some internal discussion about this, i was asked to hold on

kevoreilly:
Apologies for the delay in responding - ultimately this PR was not acceptable as it undermines one of the key design principles in cape: malware categorisation by family. The problem is that we need a single malware family name, but multiple signatures might classify the family. The design of cape is to have just the family, not the version or any other extraneous information. So for example in the case of, say, AgentTesla, the classification needs to have just this and not `AgentTeslaV3` or `AgentTeslaV4`. To solve this problem, `cape-type` was introduced to specify the family name separately to the signature name.

Thanks for the PR nonetheless, I am always open to suggestions that might improve cape. But I tried it this way in the beginning and it became a sprawling nightmare of myriad yara signature names (incidentally often with all sorts of undesirable characters or differing forms) and a lack of clear classification.

PR #1656, opened 2023-07-07
VM is frozen after analysis is complete

para0x0dise:
# Setup Environment:
I am currently utilizing physical machinery due to encountering several issues with nested virtualization using KVM. I have my own server that has equipped with Proxmox and I have installed CAPE sandbox, a FOG server responsible for VMs re-imaging to restore them to a clean state once analysis is finalized, and multiple VMs. Each machine has its own name in the FOG server and ID in Proxmox.

CAPE Design

# Main Issue: VM is frozen
By default, CAPE sandbox is configured with an analysis duration of **200 seconds**. After the analysis is finished, CAPE sends a force reboot command to the virtual machine (VM) to initiate the shutdown process. However, in some cases, the VM may encounter an issue where it remains stuck in the "Shutting Down" state without successfully completing the shutdown.

To solve this issue, we have to reset the whole machine instead of just rebooting it, so we have to communicate with the hypervisor server to make a reset request. It has several REST APIs to send and receive requests. First, I need to send my credentials to receive an access ticket, then I need to parse the ticket to get CSRF token and authentication cookie that will be sent in another request with VM-ID and action (reset request) 

This concept can be implemented on any hypervisor that provides its own set of APIs.

PROXMOX APIs


doomedraven:
Hello, love your explanations. The uniq thing that this defeats the pure FOG, which shouldn't relay on proxmon i guess. maybe add one config so in physical it could know if is pure or if is proxmon based? what do you think

doomedraven:
maybe just add an entry in physical.conf.default for something like idk, external_manager, idk not super creative for name now, and then if yes then import proxmon and use shutdown instead of direct rest api request

para0x0dise:
@doomedraven 
Done.

doomedraven:
Thank you for your job on this. I have merged proxmon configs. so you just need to put those settings in original config and thats all. can you test and if that works im fine to merge this

para0x0dise:
@doomedraven 
It worked successfully, but I modified `proxmox` identifier in `physical.py`

doomedraven:
amazing, thank you a lot for your work. merging, i forgot to remove `conf/proxmoxVMs.conf.default` can you move your conf to proxmox.conf ?

doomedraven:
i still considering to move that aux func to abstracts.py but later when i have time to think about that

PR #1659, opened 2023-07-13
az: Add support for cert auth

68zuhKQfKPk:
Adds support for alternative way for authenticating to Azure Application: by using Certificate Auth


cccs-kevin (approved):
Looks good to me, thanks!

doomedraven:
thank you both, i have removed az.conf mods as only .default should be modified

PR #1665, opened 2023-07-18
Fix exception handling for missing sudo

heck-gd:
subprocess throws FileNotFoundError when the binary does not exist.


doomedraven:
thanks

PR #1674, opened 2023-07-25
Don't lose data when multiple configs for the same family are found.

tbeadle:
Keep them separate and maintain the file_obj hashes associated with each one.

### DIFF
--- a/modules/processing/CAPE.py
+++ b/modules/processing/CAPE.py
@@ -381,9 +381,8 @@ class CAPE(Processing):
 
         # look for an existing config matching this cape_name; merge them if found
         for existing_config in self.cape["configs"]:
-            if cape_name in existing_config:
-                log.warning("CAPE: data loss may occur, existing config found for: %s", cape_name)
-                existing_config[cape_name].update(config[cape_name])
+            if cape_name in existing_config and existing_config[cape_name] == config:
+                log.info("CAPE: Found identical %s config", cape_name)
                 config = existing_config
                 break
         else:


kevoreilly:
Hmm I'm not sure about this - currently the design is that if it's a single family name, the config should be merged non-destructively. If they need to be separated they should be classified differerently.

tbeadle:
The problem we ran in to was that the original sample had a config extracted. Then a payload had a config extracted for the same family. Due to a bug in the extractor, it returned all empty strings for the fields that had been extracted from the sample, which led to data loss because those overwrote the original ones. That got me thinking that distinct configs should always just be stored separately, so that it looks like this: 
image
Note that this was taken after the bug in the config extractor was fixed and you'll notice that the payload does not have the Mutex field.

doomedraven:
looks fine to me, but i will leave this to Kevin to decide

kevoreilly:
This caters for what I perceive as a very unlikely scenario - two distinct variants from the same family with different configs in the same job. 

The issue I have is that what is a far more common scenario is that there are often multiple payloads from a single sample which may each contain the config. There are other scenarios where the whole config for a sample may be split across payloads. Finally there is the scenario in dynamic (debugger-based) config extraction where by its very nature it often involves capturing a single config item field in a dump, where the assumption is they will all be merged in the end.

It is for the above three scenarios that I had designed the current system of merging, and unfortunately to fix it for the scenario you mention of multiple variants in a single job seems to break the handling of the others which are far more common.

doomedraven:
Self update, is frequent thing

kevoreilly:
I am open to suggestions on how to handle all these cases concurrently. And although I'd be keen to keep it on by default, I would also be happy to make the merging a configurable option.

PR #1675, opened 2023-07-26
Wrong File Package

para0x0dise:
# Main Issue:
CAPE typically identifies the file type using its `CONTENT-TYPE` MIME **(as I understand)**, with assistance from the CAPE agent, and then proceeds to determine the appropriate package for executing the file. If the file type is not detected correctly, it will be identified based on its magic bytes.

Unfortunately, JAR files without their extension are treated as ZIP files. To solve this, I have modified the file type detection logic to be more specific, selecting based on the magic bytes **_ONLY_**.

filehash: 7f4e1b678de5191288c8e74a64c82d60
https://www.virustotal.com/gui/file/c55ba1aa7a57b772f0108cd8f90a36787468ef0e2a53ef9fc2bcf689d109a95c/

### DIFF
--- a/analyzer/windows/analyzer.py
+++ b/analyzer/windows/analyzer.py
@@ -331,25 +331,22 @@ class Analyzer:
         log.debug("Python path: %s", os.path.dirname(sys.executable))
 
         # If no analysis package was specified at submission, we try to select one automatically.
-        if not self.config.package:
-            log.debug("No analysis package specified, trying to detect it automagically")
+        log.debug("No analysis package specified, trying to detect it automagically")
 
-            # If the analysis target is a file, we choose the package according to the file format.
-            if self.config.category == "file":
-                package = choose_package(self.config.file_type, self.config.file_name, self.config.exports, self.target)
-            # If it's an URL, we'll just use the default Internet Explorer package.
-            else:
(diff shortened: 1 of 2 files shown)


doomedraven:
as i told packages.py is dead, and all changes should go to sflock, thanks anyway i will keep this open till i add it to sflock

doomedraven:
sflock 0.3.52 is released with proper identification, thank you

doomedraven:
i had to revert the analyzer code due to problems reported by community

para0x0dise:
Could you discuss these problems with me, please?

kevoreilly:
There's a report that shellcode packages are broken by this commit, the revert is prudent whilst we investigate, please do not take offence at this.

doomedraven:
the problem with your changes was forcing to identify package inside of the vm and totally ignore the user specified. that is not the way to go, i have missed somehow that. the `packages.py` is just there for legacy purposes but now all logic should be defined in sflock. so if you set package on submission it wasn't used and could be set to different things. is definitely not what we want. but well at least i have refreshed the understanding of that part of code. i guess we should review again packages.py ensure that everything is in packages and remove packages.py

Issue #1680, opened 2023-07-28
Python x64 versions crash the analysis process


nbargnesi:
Both the [docs](https://capev2.readthedocs.io/en/latest/installation/guest/requirements.html#install-python) and [agent module](https://github.com/kevoreilly/CAPEv2/blob/c04cc401c18c08ca5227de37fdfd4316b432c8c6/agent/agent.py#L33) say an x86 version of Python is required.

Under the covers the analysis process calls a bunch of low-level Windows libraries, unpacking the results of these calls into a series of [structures defined in lib.common.defines](https://github.com/kevoreilly/CAPEv2/blob/master/analyzer/windows/lib/common/defines.py). The analyzer assumes the structures use 32-bit sizes, and will crash if running under a 64-bit Python.

There are architecture independent ways of doing most of what the analyzer needs to do, but for now the x86 requirement is there based on how the analyzer is written.

nbargnesi:
This [commit ](https://github.com/nbargnesi/CAPEv2/commit/d41c9fec6269e594155bdb458b0077a33330fe05) fixes the struct unpacking crash when a 64-bit Python is used.

doomedraven:
can you PR this commit?

nbargnesi:
Done. I think we should keep #1680 open for some time - there will be more crashes and issues running under a x64 Python analyzer.

enzok:
The following commit implements the changes made to CAPEsolo (https://github.com/CAPESandbox/CAPEsolo) to enable 64-bit Python guests.

https://github.com/kevoreilly/CAPEv2/commit/43d5881bfdadbe2a6b8fa5f9cf9b923bcc96eec0

It still needs to be tested on a CAPE instance, if anyone has the spare cycles.

kevoreilly:
This is fixed now

PR #1686, opened 2023-08-04
partial fix for #1680, rewrite proc enum

nbargnesi:
This fixes the crash mentioned in #1680 with an architecture-agnostic strategy to process enumeration without NTDLL and struct unpacking. The backing struct was removed and the method was made into a function.


doomedraven:
thank you Nick

PR #1693, opened 2023-08-10
Set route to empty string when not specified

JSCU-CNI:
When a route is not specified in an API request to create an analysis, the route would be set to None, eventually resulting in the string 'false', which would then result in no route to be specified, instead of the default specified in the configuration settings:

https://github.com/kevoreilly/CAPEv2/blob/a0769ed88c6a92b1cb4d8422b05c2138dfe7dfe7/lib/cuckoo/core/scheduler.py#L505-L510

Setting it to an empty string if not specified resolves this issue. Setting to the string 'none' is still possible to explicitly disable routing.


doomedraven:
thank you

PR #1695, opened 2023-08-11
Fix Yara tests

josh-feather:
- Remove Redline test as rule is no longer present in the repo
- Change `test_yara()` assert to use `yara.StringMatch` to support yara 4.3.1


doomedraven:
thanks

PR #1698, opened 2023-08-11
A few things

cccs-kevin:
- `az.py`: Scale VMSS up on start if # of VMs is 0
- `routing.rst`: Correcting the routing docs
- Adding `verify_rt_table` option in `routing.conf` which affects how CAPE starts up in `startup.py`


doomedraven:
thanks, fixed config

PR #1699, opened 2023-08-12
Fix: Zipped files defaults to windows platform

kenleejl:
When submitting a zipped elf file, it goes to the windows platform. This fix will return the platform accordingly

### DIFF
--- a/lib/cuckoo/common/demux.py
+++ b/lib/cuckoo/common/demux.py
@@ -294,4 +294,4 @@ def demux_sample(filename: bytes, package: str, options: str, use_sflock: bool =
                         retlist.append(trimmed_path(filename))
                 retlist.remove(filename)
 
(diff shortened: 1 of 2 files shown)


doomedraven:
hello interesting, i think we should probably change that to set file - platform relation not only all files and platform, as if you submit different fileformats it will still be a problem

para0x0dise:
@kenleejl 
I have the same issue when submitting a python file, it goes to Linux platform

doomedraven:
this is not proper implementation as you add platform only in one place to return, that will fail on other returns. i have properly implemented this here https://github.com/kevoreilly/CAPEv2/pull/1701 should be merged on Wednesday, thanks guys for headups about this issue

doomedraven:
merged to master now

PR #1700, opened 2023-08-13
update Blister Loader yara rule

r0ny123:


kevoreilly:
Thank you

doomedraven:
this yara doens't match specified sample

kevoreilly:
Reasons for rejection:

- yara didn't match sample
- two original patterns still exist in modified form but unaltered
- one of two new patterns applies to old sample (unnecessary)
- other new pattern is the only pattern to detect new sample, so signature would become effectively a single pattern

r0ny123:
> yara didn't match sample

Partially correct. 
The mentioned sample did match the string `$xor_dec_new` (which was my intention), the problem was with the condition `2 of them`, which was why the yara didn't match. Only two of those XOR variants found ITW. I would like to have those XOR-based hex strings in YARA.

PR #1704, opened 2023-08-14
Fix mongodb too many open files error

qux-bbb:


doomedraven:
thanks, not saw this one in long time, do you have some custom setup that leaks FDs?

qux-bbb:
I should not have any custom settings that cause file descriptor leaks.  
When running many tasks, sandbox becomes stuck and MongoDB service has stopped. The mongodb log shows "too many open files" error.  
After checking cape.sh, I found that the limitNOFILE for MongoDB has been restricted.  
Maybe that is not real cause, but I can not find other error. Do you have any idea?

doomedraven:
no, in past this was a huge headache in cuckoo and supervisor. is why i moved to systemd. i didn't experienced anymore too many open files but im fine to merge, was just curious about details

PR #1708, opened 2023-08-21
Gracefully reload the maxmind db at most once every 5 minutes when it has changed

tbeadle:
Testing:
I temporarily changed the cape-processor systemd command to use `-p1` since each process.py process maintains its own latest timestamp for the geolite database. Then, after enabling country_lookup and putting the geolite country database in data/, I reprocessed a sample and verified that it loaded the country info correctly. I then reprocessed the sample and verified that it did not log that it was loading the database again. Then, I 'touch'ed the database to update the mtime and reprocessed the sample again, confirming that the 'loading maxmind database' log message was present again.

### DIFF
--- a/modules/processing/network.py
+++ b/modules/processing/network.py
@@ -23,6 +23,7 @@ from itertools import islice
 from json import loads
 from urllib.parse import urlunparse
 
+import cachetools.func
 import dns.resolver
 from dns.reversename import from_address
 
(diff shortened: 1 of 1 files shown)


doomedraven:
nice, i had something like this in mind for another things in CAPE, this a nice start for me. thanks

PR #1710, opened 2023-08-22
Bug fix to during_script.py

JLeow00:
Currently there's a bug where the sandbox will wait for the During-Execution Script to complete instead of running it concurrently with the uploaded file.

This is because of this line in the `start()` method which blocks the execution of the uploaded file until the During-Execution Script terminates

https://github.com/kevoreilly/CAPEv2/blob/1b9fd7f7a95a509c22d2ce86d7030d2a812409cc/analyzer/windows/modules/auxiliary/during_script.py#L72

The auxiliary module will work as intended if we shift the waiting and processing code to the `stop()` method


doomedraven:
@TheMythologist what do you think?

JLeow00:
Upon further testing, the sandbox won't show the stdout output of During-Execution scripts that don't explicitly flush their output or terminate gracefully, so I added a commit to disable buffering in the child process so that the output will be captured

TheMythologist:
We're still in the midst of testing the fix, but the bug is there

PR #1713, opened 2023-08-22
Added ninja-build dependency, as required by some packages

demon-i386:


doomedraven:
i would aprpeciate to know which packages requires them? as qemu bring their own version of ninja

doomedraven:
i did deploy yesterday on new servers, no issues, no need to install it, so need context

PR #1738, opened 2023-09-12
Add html scrapper auxiliary

yoniabrahamy:
- Rendering html using Selenium on the sandbox machines and uploading them back to the cape server for processing
- Cape report will have a new section -html_scrap - which will contain information extracted from the html scrap (will currently only have urls)
- Add an auxiliary module and processing module (see diagram)
- Add geckodriver as one of the auxiliary executables we send to the virtual machine to be used by selenium

image


doomedraven:
thanks, this is similar this should go to community repository, and dependency in extra/additional_dependencies.txt
also, we can't modify x.conf, ONLY x.conf.default

kevoreilly:
Thanks @yoniabrahamy this looks like an interesting feature.

From a pure language perspective I thought it might be worth pointing out that 'scrapper' is related to the verb 'to scrap' which means to throw something away as rubbish. I assume here the intended meaning is 'to scrape' websites for which the appropriate noun would be 'scraper' with a single p. Obviously this does not affect the technical output but I thought it might be worth getting the nomenclature right from the start.

yoniabrahamy:
Thanks for the feedback @doomedraven. I'll create a PR in the community repo with the auxiliary and processing module, and organize the dependency and config.

Is it ok to leave the geckodriver.exe as part of this PR?

doomedraven:
no, everything should go there as we don't need those files in core, they are not core and its pretty big file, you can leave the resultserver folder part here and `conf/processing.conf.default` part, the rest to community, but remember we don't touch `conf/processing.conf` so you need to remove those changes

doomedraven:
this looks good thanks, once the rest is merged into community i will merge this

yoniabrahamy:
Thanks @doomedraven!

The community PR is here: https://github.com/CAPESandbox/community/pull/382

doomedraven:
thanks for helping to split and move this to community

PR #1739, opened 2023-09-12
Add utility script to clone guest machines using virtsh

yoniabrahamy:
A utility script that helps with cloning guest machines on kvm using `virtsh` commands.

The script input is:

- original: The name of the virtual machine to clone from
- original-disk: The location of the disk file of the virtual machine to clone from
- prefix: The prefix for the new machine names (e.g for `prefix=win7-` the script will create `win7-1`, `win7-2` etc..)
- storage-path: The directory in which to create all of the new machine's disk files
- count: How many machine should be created
- count-offset: At what number should the count start
- ip: The base IP address machines should start to be created (e.g for base ip 192.168.122.1 the first machine will get the ip 192.168.122.1, the second machine 192.168.122.2 and so forth..)
- machine-type: which machines are being created (win7 or win10)
- dry-run: Only print out the intended work, doesn't really change anything

The script will:
- Clone the machines
- Run them for 10 mins
- Take a snapshot and shut them down
- Print out the configuration blocks to place in `kvm.conf`

### DIFF
--- a/utils/clone_machines.py
+++ b/utils/clone_machines.py
@@ -1,0 +1,281 @@
+import argparse
+import ipaddress
+import os
+import subprocess
+import tempfile
+from time import sleep
+from typing import List
+from typing import NamedTuple
+from xml.dom.minidom import parseString
+
+import randmac
+import tqdm
+
+NETWORK_NAME = 'default'
+DEFAULT_STORAGE = '/data/vm/'
+SLEEP_TIME = 650
+DEFAULT_SNAPSHOT_NAME = 'clean'
+
+
+class Machine(NamedTuple):
+    number: int
+    name: str
+    ip: str
+    mac_address: str
+    hd_path: str
+
+
+def setup_network(machines: List[Machine], is_dry_run: bool = False):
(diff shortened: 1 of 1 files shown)


cccs-kevin:
Weird, I swear this utility script used to exist in the `utils/` directory

Intezer has a version of it: https://github.com/intezer/CAPEv2/blob/c7d3ab7be334ba41667bd917cea700f9e9b334b4/utils/clone-machines.py

doomedraven:
thanks but i don't what this code in main repo, it is in community, everything that is not core related, and not maintained by core devs 

that script can be found here https://github.com/CAPESandbox/community/tree/master/utils

yoniabrahamy:
Ah I apologize @doomedraven and @cccs-kevin I didn't realize that we already provided the script to the community repo

doomedraven:
np, thats fine :) is just easier for us to separate user code from our code, as normally people drop their code and stop to update it, and in future other users just complain, so i want to have clear separation between core and not core parts. But thanks the script

PR #1740, opened 2023-09-13
Add wmplayer package

yoniabrahamy:
Some of our users uploaded mp3 files and complained the behavior report was lacking.. so we added support for it


doomedraven:
what? malware in mp3? is that for some exploit or what?

yoniabrahamy:
Yeah I agree that it is very unlikely for an mp3 file to contain malware, and I'll understand if you close this PR. 
Some of our users are sending all kinds of files without preliminary checks and just want to "see" the file running in the sandbox without harm.

doomedraven:
Os why we started to introduce filtering as we got spam of pictures on
capesandbox.com, we can't just add useless things that will waste resources
I would suggest you to start rejecting those as there is no threat

yoniabrahamy:
Very well, I'll close this PR. Thanks.

PR #1743, opened 2023-09-13
Add pdfminer integration

yoniabrahamy:
Use the [pdf miner package](https://pypi.org/project/pdfminer/) to extract more urls statically from pdf files

### DIFF
--- a/lib/cuckoo/common/integrations/pdfminer.py
+++ b/lib/cuckoo/common/integrations/pdfminer.py
@@ -1,0 +1,64 @@
+import logging
+from typing import Any
+from typing import Dict
+from typing import Iterable
+from typing import Set
+from typing import Union
+
+try:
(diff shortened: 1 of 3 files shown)


lib/cuckoo/common/integrations/pdfminer.py
```diff
@@ -0,0 +1,64 @@
...
+    from pdfminer import pdftypes
+
+    HAVE_PDFMINER = True
+except ImportError:
+    HAVE_PDFMINER = False
+
+log = logging.getLogger(__name__)
+
+
+def _search_for_url(obj: Union[pdftypes.PDFStream, dict, list]) -> Iterable[str]:
```
doomedraven, line 20: what if we move out `pdftypes.PDFStream ` from Union? to not fail on test if not installed as not all users will have it

PR #1745, opened 2023-09-13
Add feature toggle for checking path safety

yoniabrahamy:
We encountered an issue that since we store our `analyses` folder in a separate mount and create a symlink to `/opt/CAPEv2/storage`, we couldn't serve content on cape web UI since we would fail the `path_safe` function.

We added an option to toggle off this feature (default is that the feature is **enabled**)

### DIFF
--- a/conf/web.conf.default
+++ b/conf/web.conf.default
@@ -194,3 +194,7 @@ guest_rdp_port = 3389
 # Some samples will only detonate on specific versions of Windows (see web.conf packages for more info)
 # Example: MSIX - Windows >= 10
 msix = win10,win11
+
+[security]
+# When using mounted folder you might want to set to no
+check_path_safe = yes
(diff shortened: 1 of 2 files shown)


doomedraven:
as i don't have permissions to edit your PRs

```
[security]
# Useful when folder is mounted
check_path_safe = yes
```

another file, restore license header + 
```
-    if not web_cfg.security or web_cfg.security.get('check_path_safe', True):
+    if web_cfg.security.check_path_safe:
```

yoniabrahamy:
Thanks, I did the changes you suggested

doomedraven:
license header still missed

yoniabrahamy:
Right, brought it back. Sorry about that

doomedraven:
thanks, idk why i don't have permissions, otherwise i would change it by myself

PR #1753, opened 2023-09-19
Normalize files

tbeadle:
Insert the detonation-independent fields for files in to a separate collection and reference it from the analysis document.

Many of the fields stored for file objects with the same hash (target.file, dropped, CAPE.payloads, and procdump) are the same, regardless of the detonation. These can take a significant amount of space in the mongo database, so store them in a separate collection holding just those attributes once for each unique file. Keep track of the task ID's that reference them, so that they can be cleaned up later.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,15 +1,23 @@
+### [19.9.2023]
+* Storage of file data in MongoDB
+    * Store the parts of a report's file data that is independent of a detonation in a separate collection
+      to conserve disk space.
+* __ACTION REQUIRED__
+    * It is recommended to add a regular cron job to call `cd /opt/CAPEv2 && sudo -u cape poetry run python ./utils/cleaners.py --delete-unused-file-data-in-mongo`
+      to prune these entries that are no longer needed.
+
 ### [13.9.2023]
-* Monitor updates: 
+* Monitor updates:
     * .NET JIT native cache handling improvements
     * New debugger action 'string' to capture decrypted strings
     * Fix issue in procname_watch_init() with non-null-terminated unicode paths - thanks Proofpoint for the report
 
 ### [8.9.2023]
-* Monitor update: 
+* Monitor update:
     * .NET JIT native cache scanning & dumping
 
 ### [1.9.2023]
-* Monitor updates: 
+* Monitor updates:
     * Fix missing browser hooks config setting for Edge & Chrome
     * Trace: add config option to try and skip loops which flood trace logs (loopskip=1)
 
--- a/dev_utils/mongo_hooks.py
+++ b/dev_utils/mongo_hooks.py
@@ -1,0 +1,152 @@
+import itertools
+import logging
+
+from pymongo import UpdateOne
+
+from dev_utils.mongodb import (
+    mongo_bulk_write,
+    mongo_delete_data,
+    mongo_delete_many,
+    mongo_find,
+    mongo_find_one,
+    mongo_hook,
+    mongo_insert_one,
+    mongo_update_many,
+    mongo_update_one,
+)
+
+log = logging.getLogger(__name__)
+
+FILES_COLL = "files"
+FILE_KEY = "sha256"
+TASK_IDS_KEY = "_task_ids"
+
+
+def normalize_file(file_dict, task_id):
+    """Pull out the detonation-independent attributes of the given file and
+    return an UpdateOne object usable by bulk_write to upsert a
+    document into the FILES_COLL collection with its _id set to the FILE_KEY of
+    the file. The given file_dict is updated in place to remove those
+    attributes and add a 'file_ref' key containing the FILE_KEY that can be
+    used as a lookup in the FILES_COLL collection.
+    If the file has already been "normalized," then it is not modified and
+    None is returned.
+    """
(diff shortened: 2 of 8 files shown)


doomedraven:
nice stuff

stamparm:
@tbeadle @doomedraven and how to "clean them later"? what's the new workflow?

AFAICS, currently the `/apiv2/tasks/delete/<task_id>/` has changed the behavior (i.e. it is broken compared to before) as those extra collections are being filled (e.g. `files`), even though the tasks are explicitly deleted

TL;DR: before, `delete/<task_id>` deleted everything, now junk is pilled up in these separate collections

doomedraven:
yes i spot yesterday more broken stuff like find_sample_by_hash in database

tbeadle:
@stamparm 

> @tbeadle @doomedraven and how to "clean them later"? what's the new workflow?
> 
> AFAICS, currently the `/apiv2/tasks/delete/<task_id>/` has changed the behavior (i.e. it is broken compared to before) as those extra collections are being filled (e.g. `files`), even though the tasks are explicitly deleted
> 
> TL;DR: before, `delete/<task_id>` deleted everything, now junk is pilled up in these separate collections

From https://github.com/kevoreilly/CAPEv2/blob/master/changelog.md#1992023:

```
It is recommended to add a regular cron job to call cd /opt/CAPEv2 && sudo -u cape poetry run python ./utils/cleaners.py --delete-unused-file-data-in-mongo to prune these entries that are no longer needed.
```

We track which tasks have references to which file objects using the `_task_ids` key of the `files` collection. Using `cleaners.py --delete-unused-file-data-in-mongo` as described above will find ones that have no tasks that reference them any more and delete them. This is done to optimize for deleting large ranges of tasks, such as with `cleaners.py -drs ### -dre ###`. Since Mongo, AFAIK, does not have a way to update the `_task_ids` key and return its new value in the same call (like `UPDATE .... RETURNING ...` in SQL), we'd have to do a separate query after each file being deleted to see if its `_task_ids` key is empty. This could significantly slow down that operation of cleaning many tasks.

Using this asynchronous operation of cleaning those unused file objects is able to be done in 1 mongodb call that should be relatively quick, given that an index for `_task_ids` is created in startup.py.

@doomedraven can you explain more about the find_sample issue you're seeing?

stamparm:
Can I ask why there was a need for such drastic changes? Everything was
working fine, and now this seems to me like an overengineering (i.e. too
much complications for little to no gain IMHO)

doomedraven:
> @stamparm
> 
> > @tbeadle @doomedraven and how to "clean them later"? what's the new workflow?
> > AFAICS, currently the `/apiv2/tasks/delete/<task_id>/` has changed the behavior (i.e. it is broken compared to before) as those extra collections are being filled (e.g. `files`), even though the tasks are explicitly deleted
> > TL;DR: before, `delete/<task_id>` deleted everything, now junk is pilled up in these separate collections
> 
> From https://github.com/kevoreilly/CAPEv2/blob/master/changelog.md#1992023:
> 
> ```
> It is recommended to add a regular cron job to call cd /opt/CAPEv2 && sudo -u cape poetry run python ./utils/cleaners.py --delete-unused-file-data-in-mongo to prune these entries that are no longer needed.
> ```
> 
> We track which tasks have references to which file objects using the `_task_ids` key of the `files` collection. Using `cleaners.py --delete-unused-file-data-in-mongo` as described above will find ones that have no tasks that reference them any more and delete them. This is done to optimize for deleting large ranges of tasks, such as with `cleaners.py -drs ### -dre ###`. Since Mongo, AFAIK, does not have a way to update the `_task_ids` key and return its new value in the same call (like `UPDATE .... RETURNING ...` in SQL), we'd have to do a separate query after each file being deleted to see if its `_task_ids` key is empty. This could significantly slow down that operation of cleaning many tasks.
> 
> Using this asynchronous operation of cleaning those unused file objects is able to be done in 1 mongodb call that should be relatively quick, given that an index for `_task_ids` is created in startup.py.
> 
> @doomedraven can you explain more about the find_sample issue you're seeing?

it just was checking directly instead of add ref_file, my idea is to replace it with just search function from web_utils.py but i don't have free time those days to do that properly

tbeadle:
> Can I ask why there was a need for such drastic changes? Everything was working fine, and now this seems to me like an overengineering (i.e. too much complications for little to no gain IMHO)

We were regularly having detonation documents that would fail to be inserted into Mongo because they exceeded Mongo's document size limit. This was a way to reduce not only the document size but also the overall Mongo storage due to the deduplication.

stamparm:
@tbeadle you sure about "deduplication"? do you have some numbers on reduction, as to be honest, I have doubts on such statement (i.e. I expect that vast majority of files stuffed into mongodb from CAPE is fairly unique)

doomedraven:
files are uniq if you don't use a lot of `file_extra_info` features(selfextract.conf and custom modules), we do use it a lot at my $dayjob and i saw tons of duplicated files coming from different "parents"

doomedraven:
btw speaking of duplicated, we got rid of them in mongo, but we didn't get rid of them on storage. would be nice to develop solution for that too, i will ping you later tbeadle to discuss some ideas

PR #1758, opened 2023-09-21
Strip things in brackets from determination of detection based on suricata rules.

tbeadle:
Several ET suricata rules have things in brackets, like in "ET MALWARE [SEKOIA.IO] Win32/Stealc C2 Check-in". This would cause the detection to be "Sekoia" instead of "Stealc". Although there is no formal format for the msg field in rules, it seems that things in brackets can, for the most part, be disregarded.


kevoreilly:
Thank you ❤️

PR #1765, opened 2023-09-23
dl files matched by yara

doomedraven:
Useful for when you doing parser update and want to get all missed configs instead of reprocess all jobs


kevoreilly:
Lovely

PR #1769, opened 2023-09-25
Fix searching by file hashes.

tbeadle:
- For documents whose file information has been normalized to the "files"
collection, we need to search there to find the sha256 of those files
and then do our normal search using that.
- Add back searching by crc32. I think it was mistakenly taken out.
- Remove "string" from search page documentation since it was never
correctly implemented. There is no top-level "strings" field.
- Don't apply the hash suffixes to "extracted_files_tool" keys. The
corresponding value is just the name of the tool. It does not have
sub-keys for the hash types.
- Submit searches using HTTP GET so that they're easy to share. Continue
to accept POST requests, though.


kevoreilly:
Thank you

PR #1789, opened 2023-10-10
Fix typos and add Troubleshooting Documentation

SrivatsaSJoshi:
I found typos in two places while I was installing this myself so I have fixed it in this PR. I also ran into some errors that took a while to troubleshoot and that mistake can be made by any beginner so I have added how I fixed the issues in `/docs/book/src/start.rst`

Specifically, I have documented two errors:
- CuckooCriticalError : Cannot bind ResultServer on port 2042 because it was in use, bailing
- CuckooCriticalError: Unable to bind ResultServer on <IP> [Errno 99]


doomedraven:
thank you

SrivatsaSJoshi:
I didn't touch the actual code too much because I'm still not familiar with the codebase. I will change the error messages once I understand the code more.

doomedraven:
Do you know how to use tool called grep?

doomedraven:
example:

```
➜  CAPEv2 git:(staging) ✗ grep -r "Cannot bind ResultServer" .
./lib/cuckoo/core/resultserver.py:                raise CuckooCriticalError(f"Cannot bind ResultServer on port {port} because it was in use, bailing")
```

so that helps you to spot which msg from which file comes

PR #1798, opened 2023-10-17
Add package exclusion for web form

enzok:
# Remove packages from submission page. Useful when the package application is not available
# Example: packages = chrome,chromium,firefox


doomedraven:
please remove changes from `conf/web.conf` as only default should be touched

PR #1814, opened 2023-10-20
add VM machinery screenshots

nbargnesi:
Let screenshots be generated inside the VM machinery instead of inside the individual VMs.

If the [machinery_screenshots setting](https://github.com/nbargnesi/CAPEv2/blob/3e1f4d5e3ea0381ff61b1beb740b41778d349c9e/conf/cuckoo.conf#L30) is enabled, any screenshots send by VMs are discarded. This will let folks migrate to machinery screenshots without changing VMs.

On startup if the setting is enabled, the machinery is checked to make sure it supports screenshots. Startup fails if it doesn't, so we let folks correct the configuration.


nbargnesi:
similar libvirt-specific PR was also added in https://github.com/kevoreilly/CAPEv2/pull/2255

PR #1834, opened 2023-11-11
add config support for kvm-qemu.sh

VVelox:
- './kvm-config.sh' will be read in if exists now.
- various instances of '<WOOT>' have been turned into variables with default values of '<WOOT>'

This allows easier rebuilds/builds via just dumping a config file in place and then running it. No editing of the file needed.


doomedraven:
thanks

PR #1839, opened 2023-11-18
Fixed WOOT replacement bug.

t-mtsmt:
Fixed kvm-qemu.sh in a place where double quotes were missing and WOOT string replacement failed.


doomedraven:
thanks

PR #1846, opened 2023-11-24
Tentative of human like mouse movement

cccs-mog:
This specfic pattern added which happen 25% of the time, is going to induce movement of the mouse at 50 milliseconds interval for 1 second of going in roughly a straight line with fuzziness in order to be more human like. It obviously respect the rules of https://thehackernews.com/2023/11/lummac2-malware-deploys-new.html since the angle is definitely under 45 degree. The idea is for it to be generic for more than this specific implementation. Might be an addition for what is appropriate for the specific case but it does solve that too, maybe both would be good. Happy for feedbacks.


cccs-mog:
After revisiting this, it seem that it doesn't work for `b14ddf64ace0b5f0d7452be28d07355c1c6865710dbed84938e2af48ccaa46cf`. It have been corrected and it seem to work correctly now.

kevoreilly:
It does indeed work ❤️ thank you

kevoreilly:
I am happy to merge but see this was marked as draft - any reason to wait?

cccs-mog:
No reasons to wait, the reason for still being a draft is that I am working on more improvement for human.py which are related, which are still in final testing/code review but they will be on a separate PR. Can be merged without a problem !

kevoreilly:
Fantastic

doomedraven:
Thank you

PR #1847, opened 2023-11-25
Add module to parse NanoCore RAT

t-mtsmt:
I created a module to extract NanoCore RAT configuration.

The following image shows the result of extracting the configuration from this sample.  
https://bazaar.abuse.ch/sample/674427173a5e079caa90209387e6131f19ebafea2f2a0b7c580fa8ea7d4eaa45/

image


kevoreilly:
Wow this looks awesome ❤️ thanks a lot

doomedraven:
i will add tests later and then merge, have to go right now

doomedraven:
thanks for config extractor, i have added test for it + we don't use pycrypto or pycryptome, we do use pycryptodomex + small code improvements

PR #1850, opened 2023-11-28
Fixed bug in CAPEv2/installer/kvm-qemu.sh

conz27:
- a recent commit broke compilation of libvirt-glib-1.0-0
- fixed typo in generation of *.deb file
- adjusted values in /tmp/libvirt-glib_builded/DEBIAN/control
  to work as needed

### DIFF
--- a/installer/kvm-qemu.sh
+++ b/installer/kvm-qemu.sh
@@ -688,7 +688,7 @@ function install_virt_manager() {
     automake --add-missing
     ./configure
     mkdir -p /tmp/libvirt-glib_builded/DEBIAN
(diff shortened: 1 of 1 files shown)


doomedraven:
Building is open thing, but proper name and version is another as APT searches for specific thing, once I will have some spare time I will review this properly,thanks

conz27:
FWIW, I have successfully tested building CAPE several times against Ubuntu 22.04.3 with latest APT updates.

With the original implementation, `apt` was broken and was asking me to run `apt --fix-broken install`, which would have downloaded the APT version of the `libvirt-glib-1.0-0` library because the dependency for `gir1.2-libvirt-glib-1.0_1.0.0-1_amd64.deb` was not satisfied. However, once I made the changes in this PR (i.e. latest force-push), `apt` was working as expected.

PR #1856, opened 2023-11-29
Update Carbanak Config Extraction

enzok:


doomedraven:
can you add test for it plz?

doomedraven:
if you have file already can you put it here https://github.com/CAPESandbox/CAPE-TestFiles/tree/main/malware ?

PR #1867, opened 2023-12-05
Added improvement config extraction for new Nighthawk variants

Amjad50:
Some samples we have found from Nighthawk contains the profile data not in the `.profile` section, but also in another `.text` section.
I'm not sure if there are other sections used, will add more if found.

Also, improved how we handle structured config, previously it would crash if there was a dict inside a list, so this is fixed now.

### DIFF
--- a/modules/processing/parsers/CAPE/Nighthawk.py
+++ b/modules/processing/parsers/CAPE/Nighthawk.py
@@ -8,7 +8,7 @@ import regex as re
 from Cryptodome.Cipher import AES
 
 DESCRIPTION = "NightHawk C2 DLL configuration parser."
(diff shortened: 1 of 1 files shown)


doomedraven:
hello, thank you, can you please share any public hash? so we can add tests

Amjad50:
Sorry @doomedraven , at least for the moment, the sample I have is not public yet. But will add it here once it is.
But I'll see if I can find public samples with same behavior (`.text` section starts with `00` and not very large)

Amjad50:
Did this PR break carbanak somehow?

doomedraven:
thats fine, thank you anyway, no Carbanak is another think that for some reason started to break, need to investigate

PR #1879, opened 2023-12-11
Improving scaling semaphore; Adding batch tasking capability  to the scheduler

cccs-mog:
Building on https://github.com/kevoreilly/CAPEv2/pull/1757 to further address https://github.com/kevoreilly/CAPEv2/pull/1107

The active_analysis_count was also fixed in this PR.


rtkcgrantcharov:
Is this batching feature exclusive to `aws` and `az` machinery?

cccs-mog:
Hi [rtkcgrantcharov](https://github.com/rtkcgrantcharov), the batch scheduling feature should in theory work with any machinery, the reason the change are only affecting AWS and AZ is because there is no overwrite of `acquire` in other machineries. I don't know if there is a use case for support of other machineries but either way this should work (haven't been tested however)

Note that the feature of `scaling semaphore` is intended for only AWS and AZ however ( it could in theory work for any scaling machinery) as there would be no point running it for a static machinery. 

The `cuckoo.conf` file provide a description for those options. Adding this to the documentation is probably a good idea, will be working toward that.

ethhart:
@cccs-mog Just curious, with the current state of this PR, were your tasks stuck in pending?

cccs-mog:
Hi [ethhart](https://github.com/ethhart), the PR is currently in final testing by abundance of caution.

 I assume your question is regarding : https://github.com/kevoreilly/CAPEv2/pull/1879/commits/1ec0693f11a078e82ef3a84176e8d2fa4dd14c29. The idea behind that is to prevent task to be kept pending in the scenario of the machinery scaling down/breaking, which can happen. A machinery breakage is a recoverable error so a mechanism for the scheduling to handle such situations is definitely required.

If the question is more regarding the need for this PR, yes at a certain rate/number of machines the scheduler can't keep up, some machine are kept unassigned. It came from the initial discussion there: https://github.com/kevoreilly/CAPEv2/pull/1107 and is the next logical step toward a faster scheduler(having the potential for batch scheduling) after scaling semaphore for dynamic machinery.

cccs-kevin:
If anyone has any thoughts on this, let us know!

doomedraven:
looks good for me

doomedraven:
im gonna merge this to stating for 1w just in case as testing

doomedraven:
thank you for such update btw

Issue #1861, opened 2023-12-02
CapeV2 tasks don't start. Debug log shows an error 


s-ribeiro:
# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [Y] I am running the latest version
- [Y] I did read the README!
- [Y] I checked the documentation and found no answer
- [Y] I checked to make sure that this issue has not already been filed
- [Y] I'm reporting the issue to the correct repository (for multi-repository projects)
- [Y] I have read and checked all configs (with all optional parts)

# Expected Behavior

When I submit a new file for analysis the analysis should start.

# Current Behavior

After submission, the task  don't start. 

## Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions. Operating system version, bitness, installed software versions, test sample details/hash/binary (if applicable).

| Question         | Answer
|------------------|--------------------
| OS version       | Ubuntu 22.04

I've installed cape with the cape2.sh installer script. Cuckoo Sandbox 2.4-CAPE,

## Failure Logs

 
./log/cuckoo.log:2023-12-02 13:25:09,121 [lib.cuckoo.core.database] DEBUG: Database error counting tasks: The "entities" argument to Select.with_only_columns(), when referring to a sequence of items, is now passed as a series of positional elements, rather than as a list. 
./log/cuckoo.log:2023-12-02 13:25:09,123 [lib.cuckoo.core.database] DEBUG: Database error counting tasks: The "entities" argument to Select.with_only_columns(), when referring to a sequence of items, is now passed as a series of positional elements, rather than as a list. 

I could not find this issue in past issues or documentation, so I kindly request any guidance to solve it.

Thanks in advance

s-ribeiro:
Thanks.
When running the UI manually noticed that some db migrations were pending.

"sudo -u cape poetry run /usr/bin/python3 -m poetry run python manage.py migrate"

Ran this command. Restarted the cape instance and now it is working fine.

s-ribeiro:
downloaded cape2.sh 
and then executed "./cape2.sh all cape | tee cape.log"

doomedraven:
that msg is more warning than error, how did you run cape? i need more details to try to get some basic idea about your setup first

Buts94:
Same error described by @s-ribeiro. Installed CAPE with cape2.sh and executed with : "./cape2.sh all cape | tee cape.log". 
[...]

PR #1888, opened 2023-12-13
Update database.py

doomedraven:
@cccs-mog  @Buts94  can you both get a look on this? it should fix the issue + fix count


cccs-mog:
Yes seem to be fixing the issue. Wasn't having any issue with that previous syntax but like mentionned in #1861 there can be.

Behavior is as following:
COUNT(*) will count all the rows in the table, including NULL values. On the other hand, COUNT(column name) will count all the rows in the specified column while excluding NULL values. 

Already merged in so all good !

PR #1889, opened 2023-12-15
Update mongodb URL to .com to be able to download PGP key dur…

jezkerwin:
During the mongo install phase the command to download the PGP key fails with "No pgp data found"
After referring to the offical documentation is appears the URL is pgp.mongodb.com not .org


kevoreilly:
Thank you

PR #1894, opened 2023-12-18
Adding utility to download parsers, to be used with cronjob

cccs-kevin:
Addresses https://cccs.atlassian.net/browse/AL-2719

Testing in production for a bit before making PR ready for review


kevoreilly:
Curious as to what this addresses, alas I cannot access that link...

cccs-kevin:
> Curious as to what this addresses, alas I cannot access that link...

It's a method of downloading the latest parsers, without updating CAPE itself, similar to the `download_yara.py` utility

VVelox:
hmm... would not it be better to maybe move that out into the other repo?

That said I may be wrong, don't these often times require possible upgrades/updates when it comes to the depends, so may not work well? May be wrong on that bit though.

doomedraven:
that was discussed, yes they might have extra deps, so user that uses this utility uses it on his own risk :)

cccs-kevin:
Yes fair point, I'll move this utility as well as the `download_yara.py` utility to the community repo!

cccs-kevin:
Community PR: https://github.com/CAPESandbox/community/pull/390

PR #1898, opened 2023-12-19
Archive refactor

cccs-kevin:
The original goal was to just support running HTML files in archives, but then after looking at how we have multiple archive-related packages that could benefit from sharing logic, a refactor was in order.

I haven't tested the rar package or the zip_compound package, but archive and zip work

Will be running in prod for a bit before PR will be ready for review


doomedraven:
can you check please if i didn't mess on fixing conflicts?

cccs-kevin:
> can you check please if i didn't mess on fixing conflicts?

Confirmed, you are perfect 💯

doomedraven:
thanks for verification

PR #1900, opened 2023-12-21
Adding ability to try a password list for archive files

cccs-kevin:
Addresses https://cccs.atlassian.net/browse/AL-2600

We are currently able to support a single password being passed to an archive-related package. It would be great if we can just pass a list of possible passwords and the package will try all of them until it succeeds in extracting the file! This PR should do that :)

Currently only supporting Zip and Archive packages for this capability


doomedraven:
thanks

PR #1903, opened 2023-12-26
add a very basic overview of setting up Suricata based ingestion pipeline

VVelox:
Shows a basic example of integrating remote Suricata instances for sucking files off the wire and submitting them for detonation.


doomedraven:
thanks

PR #1904, opened 2023-12-28
add credits to Andrew Blair Schenck

takitakitanana:
I'm currently preparing an older project of mine to publish on GitHub, I have the same file [renamed] in my repository, and while searching for keywords in order to find and attribute the original creator, I've stumbled upon this entry with the comment of "find from where is file come and add here".

Credits added, as well as the repository where it originated from.

Best


doomedraven:
thanks

PR #1908, opened 2024-01-04
Fix AgentTesla.py

ClaudioWayne:
In some AgentTesla samples a "true" or "false" string appears after user-agent string "Mozilla/5.0..", in some it does not.
That breaks the parser/config.
Added search for filename.
Below are some examples.

---

new:

1295_fixed
---

 new:

1381_fixed
---

 new:

1408_fixed
---

 new:
1413_fixed


kevoreilly:
Thank you very much

PR #1910, opened 2024-01-04
Update cape2.sh

CarsonHrusovsky:
1.5.2 install either isn't available anymore or was having issues. 1.5.3 seems to be working.


doomedraven:
thank you, sound like we will have to make some generic check to get last version

doomedraven:
there are 1.5.4 version already
i have added automatical extraction of version `guacamole_version=$(curl -s https://downloads.apache.org/guacamole/|grep DIR|tail -1|cut -d">" -f 3| cut -d"/" -f 1)`

PR #1913, opened 2024-01-08
fixed wrong migration management

federicofantini:


doomedraven:
thank you will review that tomorrow

federicofantini:
Ok, now I've fixed the consistency problem related to the wrong applied migration.

federicofantini:
I'm not done yet, I'll pin you

doomedraven:
ok thank you

federicofantini:
@doomedraven the problem in my case is that I have the database consistent with the old "suscription" migrations but on the django model side there is a different name "subscription".

unfortunately django does not allow you to fix all possible cases, if someone has installed the project in the last two months unfortunately the third migration will be broken as there is no "subscription" field.
all the others who installed the project before will have no problems as the migration order is consistent again.

doomedraven:
interesting, i remember that there was typo fix PR sometime ago, so i guess that one generated this mess

doomedraven:
oh nice, thank you, yes sometime is pure old sql is easier than the abstraction, i will merge that to staging first as my server is running on that so i can check if all ok

PR #1914, opened 2024-01-10
Fix signature name whitelist

aslefhewqiwbepqwefbpqsciwueh:
Hi, I am hoping to get a small fix added to the whitelist check in [CAPE.py](https://github.com/kevoreilly/CAPEv2/blob/c28fcc3dfe091e47bc52d1120e4f24b39376c0de/modules/signatures/CAPE.py#L273)

The fix just normalises the process name to lower for the check. If the process name is of mixed case (eg: AcroRd32.exe), it will not hit this whitelist check. 

Many thanks for considering.


doomedraven:
thanks

kevoreilly:
Thank you

aslefhewqiwbepqwefbpqsciwueh:
Really big thank you 🙏

PR #1918, opened 2024-01-16
Scheduler signal handling improvements

josh-feather:
Improve scheduler signal handling:
- Pause/unpause the scheduler with USR1/USR2 signals
- Scheduler now waits for analyses to finish before stopping (SIGTERM)
- Updated Scheduler systemd configuration to wait 4 minutes before sending SIGKILL to allow analyses to finish


doomedraven:
Node for myself, add docs

doomedraven:
ok it will take a bit more time to merge as there are conflicts with batch scheduling, i will try to solve them this week

doomedraven:
@cccs-mog @cccs-kevin can you please also get a look here to ensure that my conflict solving doesn't break nothing

cccs-mog:
First impression is it should be good, will be testing to make sure there is no conflict. Thanks !

cccs-mog:
Line 1090 should probably have the added lines 1124 to 1126 added there as well ? There is two spot for analysis.start() in the loop. Appart from that no conflict found after testing sound good to me !

doomedraven:
added that, thank you for feedback, merging

PR #1919, opened 2024-01-17
Evented signature filtering fix

cccs-mog:
Fixed the evented signature filtering to act like a AND instead of an OR which would falsely make some signature trigger their on_call.


doomedraven:
oh thank you for work on that, i still recovering from flu so my brain was pretty off

doomedraven:
so is still draft or as latest commit say is done?

cccs-mog:
Sorry for the confusion, the commit comment done meant ready for internal code review. Should be removed from draft very soon !

doomedraven:
thanks for your work on this

cccs-mog:
Sorry took longer than expected but here it is !

doomedraven:
thank you very much

PR #1925, opened 2024-01-18
Fixing Process.py to allow running on task absent from the DB

cccs-mog:


doomedraven:
i have found few problems here, working on fix

doomedraven:
```
task["target"] = sample
TypeError: 'Task' object does not support item assignment
```

doomedraven:
is there any particular reason to change from `task.__setattr__("target", sample)` to `task["target"] = sample`? my guess is when task is none and "object" is dict not the database object right?

doomedraven:
ok then we are fine

PR #1935, opened 2024-01-23
Add auxiliary module to CAPE analyzer to log AMSI events.

tbeadle:
This (if enabled) causes an auxiliary module for the windows analyzer
called AMSICollector to run during a detonation. This assumes that the
analyzer is running as an administrator in order to have access to AMSI
events.  As each event is received, the content of the event is uploaded
to the CAPE host in "aux/amsi/<event_hash>" and metadata about the
event, such as whether or not it was detected as malicious
("scanResult") is logged to a jsonl formatted file that gets uploaded as
"aux/amsi/amsi.jsonl" when the collector is shutting down after the
detonation is complete.  The content (i.e. payload) of text-based events
is decoded from utf-16 and reencoded as utf-8 when storing the files on
the CAPE host. For non-text-based events, the .Net binary received as
the content of the event is stored as-is.  An additional processing
module has been added as well. It takes the amsi.jsonl file and stores
relevant pieces, transforming values as necessary, in MongoDB under a
new top-level key called "amsi".

This PR does not attempt to present this data to the user in any way
other than in MongoDB. That will require further work.

I understand that there is already AMSI capability in capemon, but this
pure python, auxiliary module approach enables the capturing of AMSI
events even when capemon is not loaded.


doomedraven:
thanks

PR #1937, opened 2024-01-24
Add dots for decimal separator

bartblaze:
A simple visibility / readability improvement. This is **not tested**, it should work, but would be great if someone can test!

_Example_
Will change
image

into:
image

### DIFF
--- a/web/dashboard/views.py
+++ b/web/dashboard/views.py
@@ -46,8 +46,8 @@ def index(request):
     db = Database()
 
     report = dict(
-        total_samples=f"{db.count_samples():,}",
-        total_tasks=f"{db.count_tasks():,}",
+        total_samples=f"{db.count_samples():,}".replace(',',' '),
+        total_tasks=f"{db.count_tasks():,}".replace(',',' '),
         states_count={},
         estimate_hour=None,
         estimate_day=None
--- a/web/dashboard/views.py
+++ b/web/dashboard/views.py
@@ -40,14 +40,17 @@ class conditional_login_required:
         return self.decorator(func)
 
 
+def format_number_with_space(number):
+    return f"{number:,}".replace(',',' ')
+
 @require_safe
 @conditional_login_required(login_required, settings.WEB_AUTHENTICATION)
 def index(request):
     db = Database()
 
     report = dict(
-        total_samples=f"{db.count_samples():,}".replace(',',' '),
-        total_tasks=f"{db.count_tasks():,}".replace(',',' '),
+        total_samples=format_number_with_space(db.count_samples()),
+        total_tasks=format_number_with_space(db.count_tasks()),
         states_count={},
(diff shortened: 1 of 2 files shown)


kevoreilly:
Hmm not sure if it's a country-specific thing but in the UK we do no use dots like that. That reads to me as four hundred and thirty nine point three zero four. Here we use commas for that. So the fact that it's not universal means I don't think we should adopt it.

If you are submitting PRs that are not tested, please use the staging branch not master!

doomedraven:
it easier to read i agree, in Ukraine we neither use it. but still i do like it :P

kevoreilly:
How about a comma instead of dot? Or maybe a space would be better.

cccs-kevin:
> How about a comma instead of dot? Or maybe a space would be better.

I like the "space" suggestion. Commas are used in Canada for this sort of thing

bartblaze:
Thanks for the feedback all! 😄 Spaces sound good, but not sure if Django can handle it just like that? Anyone has an idea?

doomedraven:
django can do that with custom filter, will add that in near future

nbargnesi:
> Thanks for the feedback all! 😄 Spaces sound good, but not sure if Django can handle it just like that? Anyone has an idea?

Use the l10n support in Django. I think we're at Django 4.2, so should just be a straightforward filter:

    {{ report.estimate_day | localize }}

Check out https://docs.djangoproject.com/en/4.2/topics/i18n/formatting/#std-templatefilter-localize.

Though by rough count, I see 157 templates. Localizing just one of them probably isn't ideal.

bartblaze:
> > Thanks for the feedback all! 😄 Spaces sound good, but not sure if Django can handle it just like that? Anyone has an idea?
> 
> Use the l10n support in Django. I think we're at Django 4.2, so should just be a straightforward filter:
> 
> ```
> {{ report.estimate_day | localize }}
> ```
> 
> Check out https://docs.djangoproject.com/en/4.2/topics/i18n/formatting/#std-templatefilter-localize.
> 
> Though by rough count, I see 157 templates. Localizing just one of them probably isn't ideal.

Probs better to await @doomedraven's implementation / idea, depends on the preference 😄

r0ny123:
> > How about a comma instead of dot? Or maybe a space would be better.
> 
> I like the "space" suggestion. Commas are used in Canada for this sort of thing

+1 for comma.

doomedraven:
done, here is a good explanation how to handle that with coma or make it locale https://stackoverflow.com/questions/1823058/how-to-print-a-number-using-commas-as-thousands-separators/10742904#10742904

doomedraven:
image

doomedraven:
so i have put space as is more neutral and easier to read

bartblaze:
Awesome 👍

PR #1939, opened 2024-01-26
Enhance msix package

enzok:
Handle msix packages created by Advanced Installer tool


doomedraven:
i will apply my changes on top of yours a bit later for the no args part

PR #1941, opened 2024-01-29
Construct Scoring Criteria for PE Files

para0x0dise:
I've implemented five methodologies similar to the Triage Sandbox scoring system https://tria.ge/docs/scoring:

- Score 10/10 (Malicious-Known):
Detection by YARA.

- Score [7-9]/10 (Malicious-Unknown):
If the sample triggers specific malicious signatures categories like "malware," "ransomware," "infostealer," "rat," "trojan," "rootkit," etc.
`**_Note:_** Adjusting the "weight" by 4 when detected by these categories to ensure that the finalscore will be >=7. I have also excluded any signature with confidence lower than 70%`

- Score [4-6]/10 (Suspicious-Unknown):
If the sample triggers specific suspicious signatures categories like "network", "encryption", "anti-vm", "anti-analysis", "anti-av", "anti-debug," etc.

- Score [0-3]/10 (Benign):
I had a problem here, the file hits many suspicious categories even if it's legitimate, so I've devised a temporary solution to address false positives, categorizing benign files suspiciously. By default, CAPE extracts embedded digital signatures within the PE file and verifies them using the Windows utility `signtool.exe`. If a benign file triggers any suspicious category, the system checks whether the file is digitally signed and verified.

- Score n/a (Undetected/Failed)
If the file doesn't trigger any signatures (no malicious or suspicious hit), the processtree will decide if this file is Undetected or Failed

This flowchart includes all previous cases

Malscore Criteria


doomedraven:
hey, nice works ,thank so signed malware with stolen cert automatically become benign

para0x0dise:
> hey, nice works ,thank so signed malware with stolen cert automatically become benign

Thanks for your reply, but no signed malware families will trigger at least one malicious category and will be "malicious-unknown" if no YARA hits. I made this check when the file is "suspicious-unknown"

para0x0dise:
> > hey, nice works ,thank so signed malware with stolen cert automatically become benign
> 
> Thanks for your reply, but no signed malware families will trigger at least one malicious category and will be "malicious-unknown" if there is no YARA hit. I made this check when the file is "suspicious-unknown"

para0x0dise:
> hey, nice works ,thank so signed malware with stolen cert automatically become benign

No, if malware is signed, it will activate at least one malicious category. If there is no YARA hit, it will be categorized as "malicious-unknown," and if it's a new family, we will need to perform deep analysis to either modify or create new signatures and YARA rules. I implemented this verification when the file is flagged as "suspicious-unknown."

doomedraven:
ok thanks, i will review it in next days

para0x0dise:
> thank you, it looks ok, i have merged that to staging to test that for 1w before it goes to master

Thanks a lot, I appreciate that

PR #1949, opened 2024-02-01
Flare capa 7

doomedraven:


marvinuku:
get_signatures is loaded from capa.loader now, not capa.main - otherwise works, if it helps

doomedraven:
thank you, i will update and test it, i still need to think about where to put the whole analysis summary generated by capa

doomedraven:
already in master

PR #1963, opened 2024-02-13
Use a default wait_time_to_image value of 4 for best performance

cccs-kevin:


doomedraven:
this was merged https://github.com/kevoreilly/CAPEv2/pull/1984 thank you guys as always

PR #1970, opened 2024-02-20
Fix zip_compound package

x9090:
`target_file` in `__configuration.json` expects string instead of list. Theoretically, there is no fuzzy logic needed what file is interesting as the `__configuration.json` or `target_file` option need to be specified explicitly by user.

Python 3.10 complaint `not enough values to unpack` upon returning to the caller of `process_unzipped_contents`


cccs-kevin:
@TheMythologist what do you think?

cccs-kevin (approved):

doomedraven:
thank you

PR #1975, opened 2024-02-21
Added access to the virtual session from the status.html page

CarsonHrusovsky:
This change allows users to still access their session's virtual link even if submitted via API or otherwise didn't interact with the initial virtual session link.


CarsonHrusovsky:
Ignore the initial views.py update - my repository was behind a few updates.

PR #1976, opened 2024-02-21
View remote session from status.html page

CarsonHrusovsky:
This change allows viewing the remote session from the status page as well. This is useful when submitting via API for example as the initial remote session button won't be accessible.


doomedraven:
hey, thanks, i have to improve that as for those who doesn't use guacamole they don't need to see that + there is a bug in html code, url `here` is broken

CarsonHrusovsky:
Sounds good doomed

doomedraven:
can you plz check my changes when you have oportunity?

CarsonHrusovsky:
I can look into this next week for testing - I will be away from keyboard for a bit.

doomedraven:
merged to staging for a week

PR #1980, opened 2024-02-23
Fixing AWS Deployment Bugs

plutusrt:
This will fix out of the box AWS deployment issues


doomedraven:
thank you, we can't mod `.conf`, only `.conf.default`. `.conf` will be overwritten on deployement by `.conf.default`. We do this to be able to introduce changes into .default and you can have your .conf withou conflicts on update

PR #1985, opened 2024-02-26
Update github actions to new versions

rkoumis:
Github actions showed warnings about node 16 being deprecated. Example: https://github.com/kevoreilly/CAPEv2/actions/runs/8055258257

The new versions of these actions use node 20.


doomedraven:
thank you

PR #1991, opened 2024-02-29
GitHub action refactoring

rkoumis:
- Abstract common code into a composite local action under `.github/actions`
- Document the ability to run github actions locally
- Tweak actions / tests slightly to be able to run locally

### DIFF
--- a/.github/workflows/export-requirements.yml
+++ b/.github/workflows/export-requirements.yml
@@ -9,6 +9,7 @@ on:
 
 jobs:
   update:
+    if: ${{ !github.event.act }} # skip during local actions testing
     runs-on: ubuntu-latest
     timeout-minutes: 5
     strategy:
@@ -33,6 +34,8 @@ jobs:
         run: poetry export --format requirements.txt --output requirements.txt
 
       - name: Commit changes if any
+        # Skip this step if being run by nektos/act
+        if: ${{ !env.ACT }}
         run: |
           git config user.name "GitHub Actions"
           git config user.email "action@github.com"
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -16,37 +16,13 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.10", "3.11"]
-
     steps:
-      - name: Check out repository code
-        uses: actions/checkout@v4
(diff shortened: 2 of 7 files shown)


.github/workflows/export-requirements.yml
```diff
@@ -33,6 +34,8 @@ jobs:
         run: poetry export --format requirements.txt --output requirements.txt
 
       - name: Commit changes if any
+        # Skip this step if being run by nektos/act
+        if: ${{ !env.ACT }}
```
rkoumis, line 38: The environment variable `ACT` is always set when run by `act`.

The github event `github.event.act` _may_ be set if the developer remembered to pass it in.

.github/workflows/python-package.yml
```diff
@@ -56,20 +40,24 @@ jobs:
         run: poetry run ruff . --line-length 132 --ignore E501,E402
 
       - name: Run unit tests
+        # Nektos act runs tests as root;
+        # and failing from line 10 on web/web/settings.py
+        env:
+          CAPE_AS_ROOT: 1
```
rkoumis, line 46: As mentioned in the comment, when running the github action locally, using nektos/act, the job runs as the root user.

settings.py checks to ensure we're not the root user, unless the `CAPE_AS_ROOT` environment variable is set; if not set, the tests are not running. I wanted to avoid changing settings.py.

tests/test_files.py
```diff
@@ -104,11 +104,13 @@ def test_create_invld_windows(self):
         with pytest.raises(CuckooOperationalError):
             Folders.create("Z:\\invalid\\directory")
 
+    @pytest.mark.xfail(os.geteuid() == 0, reason="nektos act runs tests as root")
     def test_delete_invld(self):
         """Test deletion of a folder we can't access."""
         dirpath = tempfile.mkdtemp()
 
         os.chmod(dirpath, 0)
+        # Deleting the directory will succeed if the tests are run as root.
```
rkoumis, line 113: When run by nektos/act and the effective userid is 0, we're able to delete the directory even when the permissions are 0.  So it won't raise the expected CuckooOperationalError.

doomedraven:
thank you a lot on helping improving this

PR #1995, opened 2024-03-01
Better control of environment variable for nektos act

rkoumis:
- Moved the setting of the `CAPE_AS_ROOT` environment variable to a new `.actrc` file
  - This is automatically applied when running nektos act, and not otherwise
- Removed the setting of the `CAPE_AS_ROOT` env.var from the github workflow definition files
- Updated the documentation


doomedraven:
thank you

PR #1996, opened 2024-03-02
public and private package configuration

nbargnesi:
This adds two new methods to [lib.common.abstracts.Package](https://github.com/kevoreilly/CAPEv2/blob/76754ee2d76e136285d85234ccdab79b3ff1e6e8/analyzer/windows/lib/common/abstracts.py#L21):
1. [configure(target)](https://github.com/nbargnesi/CAPEv2/blob/8f8c4b08c7105f8c5fe11c2dec15c899235fbf30/analyzer/windows/lib/common/abstracts.py#L56)
2. [configure_from_data(target)](https://github.com/nbargnesi/CAPEv2/blob/8f8c4b08c7105f8c5fe11c2dec15c899235fbf30/analyzer/windows/lib/common/abstracts.py#L70)

Both methods are optional.

If a package doesn't provide a configure method, it's [logged but ignored](https://github.com/nbargnesi/CAPEv2/blob/8f8c4b08c7105f8c5fe11c2dec15c899235fbf30/analyzer/windows/analyzer.py#L563). Likewise if no private configuration is importable from `data.packages.<package_name>`, it's also [logged but ignored](https://github.com/nbargnesi/CAPEv2/blob/8f8c4b08c7105f8c5fe11c2dec15c899235fbf30/analyzer/windows/analyzer.py#L573).

With this PR, packages like [DOC_ANTIVM](https://github.com/kevoreilly/CAPEv2/blob/master/analyzer/windows/modules/packages/doc_antivm.py) can separate 40 lines of configuration into a separate method, and provide a way to mask sensitive configuration.

### DIFF
--- a/analyzer/windows/analyzer.py
+++ b/analyzer/windows/analyzer.py
@@ -556,6 +556,27 @@ class Analyzer:
         subprocess.call("net start winmgmt", startupinfo=si)
         # log.info("Started WMI Service")
 
+        # Do any analysis package configuration
+        try:
+            self.package.configure(self.target)
+        except NotImplementedError:
+            # let it go, not every package is configurable
+            log.debug("package %s does not support configure, ignoring", package_name)
+        except Exception as e:
+            raise CuckooPackageError("error configuring package %s: %s", package_name, e) from e
+
+        # Do any package configuration stored in "data/packages/<package_name>"
+        try:
+            self.package.configure_from_data(self.package, self.target)
+        except ModuleNotFoundError:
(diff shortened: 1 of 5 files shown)


doomedraven:
for what do you need the name of the package?

nbargnesi:
> for what do you need the name of the package?

Ah, that was from an earlier version of this where I had the package classes calling `configure_from_data`. That's not needed, I'll pull it.

doomedraven:
nice feature thank you

PR #1999, opened 2024-03-06
Run ruff during pre-commit checks

rkoumis:
I frequently get `ruff` finding issues and this causes CI jobs to abort.

Would this be an improvement? I am not completely sure.


doomedraven:
thank you for all your work on this

PR #2000, opened 2024-03-07
Signature segregation

winson0123:
Splitting of signature processing for Windows and Linux. Separated to include an "all" categorization which will run common/static signaturization(e.g. network, strings, etc) for all OSes.


doomedraven:
thank you very much

PR #2001, opened 2024-03-07
Linux support

winson0123:
Long awaited Linux support.

Description of how behavioral information is captured by Linux Machines:
- [strace](https://strace.io/) is used to detonate the malware sample, and the logs are captured and uploaded to the host.
- Strace logs are processed and outputted similarly to how the current implementation of windows behaviors are. 
- Strace information is stored onto the "behaviors" key in the CAPE report json (like how Windows would)
- Information is fetched and retrieved from the database and is displayed on the behavioral tab
- Behavioral tab showcases all the syscalls that are called by each process that is spawned by the sample
- Syscalls are filtered by the root of the kernel source code (`mm`, `fs`, `crypto`, etc)
- Small hiccup we've encountered running ransomware samples, it encrypts the strace log and prevents us from processing the logs for behavioral analysis

Database changes:
- Adds the column `platform` to the `guests` table in PostgreSQL
- Required to run alembic migrations

Future work:
- More signaturization for Linux samples is possible


doomedraven:
hello amazing job on this, i got first quick look, i will check properly a bit later. but so far looks good, i have a question, to prevent ransom encrypt logs, maybe we should stream the strace log to the server side? it sound to me that we had something for that i will have to double check

nbargnesi:
And well-timed with the ability to run tests against the analyzer and agent! (CC @rkoumis )

analyzer/linux/lib/core/packages.py
```diff
@@ -157,38 +158,55 @@ def finish(self):
...
+    #     )
+
+    #     while b"systemtap_module_init() returned 0" not in self.proc.stderr.readline():
+    #         # log.debug(self.proc.stderr.readline())
+    #         pass
+
+    #     stap_stop = timeit.default_timer()
+    #     log.info("Process startup took %.2f seconds", stap_stop - stap_start)
+    #     return True
+
```
nbargnesi, line 195: probably worth gutting any systemtap-based support in the analyzer at this point

winson0123:
> hello amazing job on this, i got first quick look, i will check properly a bit later. but so far looks good, i have a question, to prevent ransom encrypt logs, maybe we should stream the strace log to the server side? it sound to me that we had something for that i will have to double check

I'll probably take a look into this next week, that suggestion does sound better to circumvent the encryption issue. Probably something along the lines of piping the output on the agent and streaming to the server side.

analyzer/linux/lib/core/packages.py
winson0123: Even the auxiliary module [stap.py](https://github.com/kevoreilly/CAPEv2/blob/master/analyzer/linux/modules/auxiliary/stap.py)?

kevoreilly:
Thank you ❤️

analyzer/linux/lib/core/packages.py
nbargnesi: That looks great!

Years ago SystemTap would have been a go-to tool for a lot of use cases, though it had its limitations. In the meantime eBPF has taken over, replacing nearly every use case I can think of for SystemTap, and doing it better.

doomedraven:
i have one question, is not related to fully related to this PR, but as we nuke stap, we need to remove it from the `.conf` maybe we should already nuke all not `.conf.default` as we touch already one we can delete directly it maybe as it will be overwritten anyway on install, any thoughts to help me decide the proper way?

PR #2002, opened 2024-03-08
Automated unit tests for agent; monitoring of async python process; input validation when setting status; added support for mutexes

rkoumis:
All these changes are backwards-compatible; in other words, there are
no breaking changes.

Agent version 13. Unit tests for agent
==
- Added file test_agent.py
- Tests will be run in Windows and Linux
- Tests will be run in github actions
- Test most existing functionality of the agent
- In send_file open the file in binary mode (bug fix)
- Updates to the agent to make it testable, including:
  - Pass a multiprocessing event to the run() method when under test, so the test knows when the agent process is ready
  - Tweaks to the shutdown method enabling testing
- Let jsonify not crash if values cannot be serialized
- Add a new command-line parameter, -v, useful when testing interactively
  - When -v is given, stdout and stderr will simply go to the console
- Allow the 'date' command to be executed from localhost; for testing

Agent version 14. Stricter checks when setting agent status
==
- Use an enum for the status. Only accept expected values.
- Check that we can read a file before trying to send it.

Agent version 15. Monitor a background process
==
- Monitor async python process spawned in background
  - Be able to detect if background process completed ok or errored
- Accurately report failure status on execution failures
  - The agent used to report RUNNING when the process actually FAILED
- Add base64 encoding capability to send_file
- Detect and log errors that occur during send_file
- Allow json_success to have status codes
- Allow json_error to accept kwargs, like json_success already does
- More detailed error messages for certain kinds of failure
  - creating directory; storing file; extracting zip file

Agent version 16. Added support for locking and releasing mutexes
==
- This feature only available in Windows
- This can serve a variety of purposes
- Via a POST request, the agent will open the named mutex
  - If not immediately available, wait 500 ms
- Via a DELETE request, the agent will release the named mutex
- The mutex must already exist; the agent will not create a mutex

### DIFF
--- a/.github/workflows/python-package-windows.yml
+++ b/.github/workflows/python-package-windows.yml
@@ -23,14 +23,21 @@ jobs:
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5
+        # Use x86 python because of https://github.com/kevoreilly/CAPEv2/issues/168
         with:
           python-version: ${{ matrix.python-version }}
           cache: 'pip'
+          architecture: 'x86'
 
-      - name: Install pytest
-        run: pip install pytest
+      - name: Install dependencies
+        run: pip install --upgrade pytest requests
 
-      - name: Run unit tests
+      - name: Run analyzer unit tests
         run: |
           cd analyzer/windows
+          pytest -v .
+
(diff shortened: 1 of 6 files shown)


doomedraven:
wow that a great work, thank you

PR #2005, opened 2024-03-11
Improvement to process.py logging

rkoumis:
Log the image name as well as the process ID; added tests. 
Also, restore the process is_alive check, disabled 4 years ago.

Credit: @nbargnesi

Sample log entries:
```
2024-03-10 12:12:13,411 [lib.api.process] DEBUG: getting exit code for <Process 17492 aspnet_compiler.exe>
2024-03-10 12:12:13,411 [lib.api.process] DEBUG: <Process 17492 aspnet_compiler.exe> exit code is 259
2024-03-10 12:12:13,411 [lib.api.process] DEBUG: <Process 17492 aspnet_compiler.exe> is still active
```

### DIFF
--- a/analyzer/windows/lib/api/process.py
+++ b/analyzer/windows/lib/api/process.py
@@ -13,48 +13,54 @@ import sys
 import urllib.error
 import urllib.parse
 import urllib.request
-from ctypes import byref, c_int, c_ulong, create_string_buffer, sizeof
+from ctypes import byref, c_buffer, c_int, c_ulong, create_string_buffer, sizeof
 from pathlib import Path
 from shutil import copy
 
-from lib.common.constants import (
-    CAPEMON32_NAME,
-    CAPEMON64_NAME,
-    LOADER32_NAME,
-    LOADER64_NAME,
-    LOGSERVER_PREFIX,
-    PATHS,
-    PIPE,
-    SHUTDOWN_MUTEX,
-    TERMINATE_EVENT,
-)
 from lib.common.defines import (
     CREATE_NEW_CONSOLE,
     CREATE_SUSPENDED,
     EVENT_MODIFY_STATE,
     GENERIC_READ,
     GENERIC_WRITE,
-    KERNEL32,
-    NTDLL,
+    MAX_PATH,
     OPEN_EXISTING,
     PROCESS_ALL_ACCESS,
     PROCESS_INFORMATION,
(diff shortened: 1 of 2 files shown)


doomedraven:
LGTM, what about you @kevoreilly ?

kevoreilly:
If we can quiten these messages down to an 'acceptable' level then I think it should be ok.

rkoumis:
I've commented out the verbose / spammy logging, @kevoreilly @doomedraven 

I squashed the commits back down to one, and now I realize I probably shouldn't have, sorry. Anyhow, that's what I did. Let me know if any questions.

doomedraven:
im not having time right now to test, was you able to test @kevoreilly latest commit? if no i will do that in 1-2 days

kevoreilly:
Yep I'll test it tomorrow

rkoumis:
no rush but please take another look when you get a chance

kevoreilly:
I've been running this on my main server for the last few days - I just need to check the logs.

kevoreilly:
I wonder if this line is ok:

```
log.debug("Failed getting exit code for %s", self)
```

Is ``self`` not gonna bomb as not a string type?

analyzer/windows/lib/api/process.py
```diff
@@ -642,6 +682,11 @@ def upload_memdump(self):
...
             log.error(file_path)
-        log.info("Memory dump of process %d uploaded", self.pid)
+        log.info("memory dump of %s uploaded", self)
 
         return True
+
+    def __str__(self):
+        """Get a string representation of this process."""
+        image_name = self.get_image_name() or "???"
+        return f"<{self.__class__.__name__} {self.pid} {image_name}>"
```
rkoumis, line 692: This method allows us to do `log.info("%s", self)`

kevoreilly: Ah ok cool - sorry to be a pain!

rkoumis:
The new `__str__()` method will take care of converting a `Process` to a string. That's how come in your first round of testing you saw messages like
```
DEBUG: <Process 4364 explorer.exe> is still active
```

```python
    def __str__(self):
        """Get a string representation of this process."""
        image_name = self.get_image_name() or "???"
        return f"<{self.__class__.__name__} {self.pid} {image_name}>"
```

kevoreilly:
Thank you!

PR #2006, opened 2024-03-12
Omit setting the length when in base64 transmission mode.

rkoumis:
Omit setting the length when in base64 transmission mode.

It was setting the file length, which was the wrong length for the encoded data.
The received data were truncated.

### DIFF
--- a/agent/agent.py
+++ b/agent/agent.py
@@ -40,7 +40,7 @@ if sys.version_info[:2] < (3, 6):
 if sys.maxsize > 2**32 and sys.platform == "win32":
     sys.exit("You should install python3 x86! not x64")
 
-AGENT_VERSION = "0.16"
+AGENT_VERSION = "0.17"
 AGENT_FEATURES = [
     "execpy",
     "execute",
@@ -49,6 +49,8 @@ AGENT_FEATURES = [
     "largefile",
     "unicodepath",
 ]
+BASE_64_ENCODING = "base64"
+
 if sys.platform == "win32":
     AGENT_FEATURES.append("mutex")
     MUTEX_TIMEOUT_MS = 500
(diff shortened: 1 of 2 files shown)


agent/agent.py
```diff
@@ -257,26 +259,30 @@ class send_file:
...
 
+    def okay_to_send(self):
+        return os.path.isfile(self.path) and os.access(self.path, os.R_OK)
+
     def init(self):
-        if os.path.isfile(self.path) and os.access(self.path, os.R_OK):
-            self.length = os.path.getsize(self.path)
+        if self.okay_to_send():
+            if self.encoding != BASE_64_ENCODING:
+                self.length = os.path.getsize(self.path)
```
rkoumis, line 273: This is the fix.  Previously we were always setting the length.  Now, we'll skip setting the length when base64 transmission is called for.

agent/agent.py
```diff
@@ -257,26 +259,30 @@ class send_file:
...
         else:
             self.status_code = 404
-            self.length = 0
 
     def write(self, httplog, sock):
-        if not self.length:
+        if not self.okay_to_send():
```
rkoumis, line 278: Can't use self.length any longer to check if it's okay to send the file.

agent/test_agent.py
```diff
@@ -473,6 +473,14 @@ def test_retrieve(self):
...
         assert first_line in r.text
         assert last_line in r.text
+        # Also test the base64-encoded retrieval.
+        form["encoding"] = "base64"
+        r = requests.post(f"{BASE_URL}/retrieve", data=form)
+        assert r.status_code == 200
+        decoded = base64.b64decode(r.text + "==").decode()
+        assert "test data" in decoded
+        assert first_line in decoded
+        assert last_line in decoded
```
rkoumis, line 483: Make sure we receive even the very last line of the file.

doomedraven:
thanks

PR #2010, opened 2024-03-13
fixed unrecognized bytes encoding in remcos parser

federicofantini:
Hi, some days ago I noticed this error due to undecodable data in the ioc extractor config tab (`/apiv2/tasks/get/config/task_id/`): `UnicodeDecodeError: 'utf-8' codec can't decode byte 0x81 in position 1: invalid start byte`

The content of the keys with index 54 and 55 are similar to random bytes and DRF fails to display it because it tries to decode the content with the default utf8 encoding.

As with the following keys I have encoded the content as base64 to be able to display it in the view.


kevoreilly:
Thanks

PR #2011, opened 2024-03-13
Rozena/Swort (Metasploit) payload parser

para0x0dise:
During our Threat-Hunting process, we noticed many samples tagged with Rozena/Swort payloads, which are meterpreter reverse shell payloads; these payloads are encoded with a well-known encoder called **"Shikata Ga Nai"**.
The decoding process basically depends on XOR operations with hard-coded 4 bytes and could be 1 encoded block or many. I tried to decode it statically, but these blocks have no reliable pattern to analyze, so I have used **Unicorn** emulator to decode it. The payload itself uses [Dynamic API Resolution/API hashing](https://attack.mitre.org/techniques/T1027/007/) to call `connect` function with `IP:PORT` in runtime, so it's easy to find a common pattern of where it pushed the IP and the Port onto the stack. This should work fine with 32-bit files.

I have tested it on the following hashes:
0254ed6c06048d7094531ed34a066e0a03ab67d5cc4beace0dc5efa8a9a1d982
034cf2b723753fd401b50906f222ce4933a811fe2625cb1d291f35e80dc7131c
072b699a7863d018f0366b300c96bea4cbd9a4f13a99e8ca99321014e2911160
094f051060f77d4ce45b00397dbc9c21632d43b59f4f7e7b161fb41a6d6ebcb6
0a35d2dc8c47c245ff096993387e16a2f43fc83a33f26db47e2fe952ad38ce45
0b5132488018b6644dc10ab00525c714aa03b13bc48a615e4981ba4811263c23
1662df2a6a4801b705994cba1d6ea31e29807c551fb28555e355c70a65f3271a
1cec5408893be1b8942c9164c41df9dea46494aa06a998ab1892354a112d6e00
28eef580a9fa7e0b98491022d6f07c4ef59971ace90b994b4fbf52fa173d04f3
3358536b67259eecd39ea7c7e17b2c7b83731ab87f89171837e863e2a2ecce53
40919a5260caa74b183953fb8dc34e59e23d9e6dc7b1878dff67caf3b0f93230
73e4c3bcc060ca9bc5cb9a13201029a03ae5fd131b84261a9ad8fbc59e22dac5
7ec18aef23ca2381313a2d4c9bd46729c5d6ae81c580bd442483ac7a2d2c8257
a74a71ef2cef0dc90b066986cc022094154c1fe09e0bbdbcaefb0e30082207ee
aeb05df0b865df62cebceab0e5259d69d39a1c8d4f623a2491dc302ac56ee9f0
b01651bfef7908d1f51b301aae0104711307be6c05b8384cb1167f2daeff520d
b04231bab6484665dda078815ad1affb339959ffec4f14f8a7ddd6b5b6df5a1e
bb12c5ce83970ae19ff59f146882e16f2911cd4fd43053d934e78d19c5327291
c32ddf1ce6946115e2c3515299997d7263601a9c3ea0313eadf9f82c4c59d8f0
da0acd72e3a321797a8d9c1a77b0aedcb8b8e0d675d889eec2911b5550051f4c
ee5af9b216c28c18011f644f4bffbd1bfb045f69b36b9fa1215b56831294d4f9
f21ef6547b4c6099c7cf8c0efd569ed527d7919a7e7446753a49c623f38e27c3

### DIFF
--- a/data/yara/CAPE/Rozena.yar
+++ b/data/yara/CAPE/Rozena.yar
@@ -1,0 +1,35 @@
+rule Metasploit_Payload {
+    meta:
+        author    = "Steven Miller"
+        company   = "FireEye"
+        reference = "https://www.fireeye.com/blog/threat-research/2019/10/shikata-ga-nai-encoder-still-going-strong.html"
+        modified_by = "Mohannad Raafat"
+        cape_type = "Rozena Payload"
+
+    strings:
+        // CASE-1: If the shellcode starts with moving the XOR Key
+        $varInitializeAndXorCondition1_XorEAX = { B8 ?? ?? ?? ?? [0-30] D9 74 24 F4 [0-10] ( 59 | 5A | 5B | 5C | 5D | 5E | 5F ) [0-50] 31 ( 40 | 41 | 42 | 43 | 45 | 46 | 47 ) ?? }
+        $varInitializeAndXorCondition1_XorEBP = { BD ?? ?? ?? ?? [0-30] D9 74 24 F4 [0-10] ( 58 | 59 | 5A | 5B | 5C | 5E | 5F ) [0-50] 31 ( 68 | 69 | 6A | 6B | 6D | 6E | 6F ) ?? }
+        $varInitializeAndXorCondition1_XorEBX = { BB ?? ?? ?? ?? [0-30] D9 74 24 F4 [0-10] ( 58 | 59 | 5A | 5C | 5D | 5E | 5F ) [0-50] 31 ( 58 | 59 | 5A | 5B | 5D | 5E | 5F ) ?? }
+        $varInitializeAndXorCondition1_XorECX = { B9 ?? ?? ?? ?? [0-30] D9 74 24 F4 [0-10] ( 58 | 5A | 5B | 5C | 5D | 5E | 5F ) [0-50] 31 ( 48 | 49 | 4A | 4B | 4D | 4E | 4F ) ?? }
+        $varInitializeAndXorCondition1_XorEDI = { BF ?? ?? ?? ?? [0-30] D9 74 24 F4 [0-10] ( 58 | 59 | 5A | 5B | 5C | 5D | 5E ) [0-50] 31 ( 78 | 79 | 7A | 7B | 7D | 7E | 7F ) ?? }
+        $varInitializeAndXorCondition1_XorEDX = { BA ?? ?? ?? ?? [0-30] D9 74 24 F4 [0-10] ( 58 | 59 | 5B | 5C | 5D | 5E | 5F ) [0-50] 31 ( 50 | 51 | 52 | 53 | 55 | 56 | 57 ) ?? }
+
+        // CASE-2: If the shellcode starts with 'fcmovu' instruction and 'fnstenv'
+        $varInitializeAndXorCondition2_XorEAX = { (DA | DB | DC | DD) ?? D9 74 24 F4 [0-30] B8 ?? ?? ?? ?? [0-10] ( 59 | 5A | 5B | 5C | 5D | 5E | 5F ) [0-50] 31 ( 40 | 41 | 42 | 43 | 45 | 46 | 47 ) ?? }
+        $varInitializeAndXorCondition2_XorEBP = { (DA | DB | DC | DD) ?? D9 74 24 F4 [0-30] BD ?? ?? ?? ?? [0-10] ( 58 | 59 | 5A | 5B | 5C | 5E | 5F ) [0-50] 31 ( 68 | 69 | 6A | 6B | 6D | 6E | 6F ) ?? }
+        $varInitializeAndXorCondition2_XorEBX = { (DA | DB | DC | DD) ?? D9 74 24 F4 [0-30] BB ?? ?? ?? ?? [0-10] ( 58 | 59 | 5A | 5C | 5D | 5E | 5F ) [0-50] 31 ( 58 | 59 | 5A | 5B | 5D | 5E | 5F ) ?? }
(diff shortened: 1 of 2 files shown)


kevoreilly:
Hi para0x0dise - firstly thank you for submitting this PR and and the work involved. As you probably know, since its inception cape has always had at its core a strong focus on dynamic extraction of 'payloads'. Therefore it may not surprise you to hear that the first thing I thought when I read your PR is that there must be a bug in cape. Why would we need to use emulation to decode something statically which is surely something cape should be capturing dynamically?!

So whilst I admire your solution to use emulation to solve this problem, there was actually a bug/issue in cape where the shellcode payload was not captured (example ``3358536b67259eecd39ea7c7e17b2c7b83731ab87f89171837e863e2a2ecce53``:

```
2024-03-14 10:43:44,220 [root] DEBUG: 2296: caller_dispatch: Added region at 0x00420000 to tracked regions list (ntdll::LdrLoadDll returns to 0x005BE55A, thread 404).
2024-03-14 10:43:44,220 [root] DEBUG: 2296: ReverseScanForNonZero: Error - Supplied address inaccessible: 0x0041FFFF
```

I was able to resolve this issue with a refinement to capemon: https://github.com/kevoreilly/capemon/commit/7dc3a4dae2c300cff4c5b50415e1da731cb0256e

This results in the decrypted shellcode being captured as a 'payload' in line with cape's core design:

image

As you can see, this payload is already detected by an existing signature which is ultimately the same signature by Steven Miller of FireEye and can be found in the community repository in ``data\yara\binaries\shellcodes.yar`` as ``shellcode_shikataganai_encoding``.

I have pushed the updated monitor in https://github.com/kevoreilly/CAPEv2/commit/2b9b122110856a9e5703c6b94757597f41a6d8bd - please test this update to ensure it captures all the payloads you are interested in. Due to the dynamic mechanism this should theoretically include similar samples not covered by the yara signature. With this fix and the existing detection, hopefully this PR is unnecessary (whilst of course gratefully received!).

kevoreilly:
The parser to grab the IP address and port could of course still be useful - for this I think it would be better to use detection of the actual decoded code rather than the decoder. Something like:

```
rule Rozena
{
    meta:
        cape_type = "Rozena Payload"
    strings:
        $ip_port = {FF D5 6A 0A 68 [4] 68 [4] 89 E6 50 50 50 50 40 50 40 50 68 [4] FF D5}
    condition:
        all of them
}
```

... for example. This would obviously benefit from a few more choice functional code blocks.

para0x0dise:
Hello @kevoreilly, thanks a lot for the great tips, but unfortunately, I have re-tested CAPEMON, and it successfully extracted the shellcode on the following hashes only:
3358536b67259eecd39ea7c7e17b2c7b83731ab87f89171837e863e2a2ecce53
7ec18aef23ca2381313a2d4c9bd46729c5d6ae81c580bd442483ac7a2d2c8257
40919a5260caa74b183953fb8dc34e59e23d9e6dc7b1878dff67caf3b0f93230

kevoreilly:
Ok thanks I'll try to see what's up with the others.

kevoreilly:
Damn. The others are falling foul of filters I added in the last few months to try and solve the problem of excessive payloads.

```
	if (TrackedRegion->SubAllocation && Size < SystemInfo.dwPageSize)
	{
		DebugOutput("ProcessTrackedRegion: Sub-allocation at 0x%p skipped due to size 0x%x", Address, Size);
		TrackedRegion->PagesDumped = TRUE;
		return;
	}
```

Here the shellcode is both a 'sub-allocation' and its size is a fair bit smaller than the page size.

```
2024-03-14 21:32:11,096 [root] DEBUG: 1356: ProcessTrackedRegion: Sub-allocation at 0x02280000 skipped due to size 0x2f7
```

I will have to try and refine or scrap this filter.

kevoreilly:
Right I am happy to say I found a far better way to filter the unwanted .NET payloads that prompted me to add this terrible size filter a few months ago https://github.com/kevoreilly/capemon/commit/4daff8c7857e2ee9278f39a8a81f9a9e24e22651

So now all the payloads should be captured! Please test the latest update https://github.com/kevoreilly/CAPEv2/commit/9616452c936bf7c7dce2261ce3fdeb90af3ef5d7 and let me know if this results in success for all your samples. Thanks for your help and patience in getting to the bottom of this issue.

para0x0dise:
Incredible, now, CAPE could detect more samples, but there is still a problem in the following hashes:
 73e4c3bcc060ca9bc5cb9a13201029a03ae5fd131b84261a9ad8fbc59e22dac5
 a74a71ef2cef0dc90b066986cc022094154c1fe09e0bbdbcaefb0e30082207ee
aeb05df0b865df62cebceab0e5259d69d39a1c8d4f623a2491dc302ac56ee9f0
da0acd72e3a321797a8d9c1a77b0aedcb8b8e0d675d889eec2911b5550051f4c
c32ddf1ce6946115e2c3515299997d7263601a9c3ea0313eadf9f82c4c59d8f0
28eef580a9fa7e0b98491022d6f07c4ef59971ace90b994b4fbf52fa173d04f3

para0x0dise:
> Hmm you are right these samples do not seem to run at all.... This doesn't appear to be related to any recent changes, looks like these don't run on any monitor version.... or even without a monitor! They appear to not even run on any.run! https://app.any.run/tasks/a0ea228f-8a33-48c1-8fab-987ffc81b26f
> 
> So I think these particular samples might be duff...

Totally agreed, but here, it seems to be run successfully on other sandboxes in VirusTotal (it could be run once detected and now it's dead). The question is, what if we need to extract important configs from a dead sample?
image

kevoreilly:
I think we need to get to the bottom of whether it can run or not... because if it can then it needs to in cape.

kevoreilly:
From what I can tell it looks to have bombed in 'zenbox' and 'sysinternals' too due to the only behavioral output being associated with werfault.

kevoreilly:
Ultimately if you can decode these statically then I suppose there is no harm in having a parser that does so.

para0x0dise:
> Ultimately if you can decode these statically then I suppose there is no harm in having a parser that does so.

Hmm, may be in another PR, but I think we are good with new results. Thanks a lot for a great work.

kevoreilly:
Thank you ❤️

PR #2021, opened 2024-03-20
Fix issues with looking for specific config files

tbeadle:
There might not be a conf/cuckoo.conf file, i.e. if the user has simply created custom/conf/cuckoo.conf to override the defaults.


doomedraven:
thanks

PR #2030, opened 2024-03-28
Update kvm-qemu.sh

qux-bbb:
Change "apt" to "apt-get".


doomedraven:
totally forgot about this one, thanks

PR #2031, opened 2024-03-28
Fix for the scheduling problem on non batch scheduling

cccs-mog:


cccs-mog:
It fail the test because the function acquire_machine now look for scheduled machines only when assigning. Need to add this to the test.

doomedraven:
thanks for the logic fix

PR #2032, opened 2024-03-29
Skip subprocess call if user STOP is sent.

azurda:
state.get("async_subprocess") is resetting back the status to RUNNING even if it was forcefully changed to COMPLETE.


doomedraven:
thanks

PR #2036, opened 2024-03-30
Fix process log

qux-bbb:
I found that the process.log doesn't have log, and each task's log such as process-123.log doesn't have log.  
Now I know the process.log should have log. The autoprocess has "log.info", but the message cannot be logged.    
It's normal that the process-123.log doesn't have log, because there is no "log.info" in process func.  

I fix this bug by separating init_logging func.

### DIFF
--- a/utils/process.py
+++ b/utils/process.py
@@ -109,7 +109,7 @@ def process(
         main_task_id = get_options(task_dict["options"]).get("main_task_id", 0)
 
     # ToDo new logger here
-    handlers = init_logging(tid=str(task_id), debug=debug)
+    per_analysis_handler = init_per_analysis_logging(tid=str(task_id), debug=debug)
     set_formatter_fmt(task_id, main_task_id)
     setproctitle(f"{original_proctitle} [Task {task_id}]")
     results = {"statistics": {"processing": [], "signatures": [], "reporting": []}}
@@ -154,10 +154,7 @@ def process(
         for i, obj in enumerate(gc.garbage):
             log.info("(garbage) GC object #%d: type=%s", i, type(obj).__name__)
 
-    for handler in handlers:
-        if not handler:
-            continue
-        log.removeHandler(handler)
+    log.removeHandler(per_analysis_handler)
 
 
 def init_worker():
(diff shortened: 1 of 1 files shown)


doomedraven:
thanks, i will check this on tuesday

qux-bbb:
Is there any problem?

doomedraven:
lack of time

doomedraven:
sorry for delay, had another priorities to handle first.

the log is done to be in logs or in analysis folder
```
if logconf.logger.process_analysis_folder:
            path = os.path.join(CUCKOO_ROOT, "storage", "analyses", str(tid), "process.log")
        else:
            path = os.path.join(CUCKOO_ROOT, "log", "process-%s.log" % str(tid))
```

you neither has it there? as it works for me just fine

qux-bbb:
> sorry for delay, had another priorities to handle first.
> 
> the log is done to be in logs or in analysis folder
> 
> ```
> if logconf.logger.process_analysis_folder:
>             path = os.path.join(CUCKOO_ROOT, "storage", "analyses", str(tid), "process.log")
>         else:
>             path = os.path.join(CUCKOO_ROOT, "log", "process-%s.log" % str(tid))
> ```
> 
> you neither has it there? as it works for me just fine

I have those codes， but there is a problem.  
Before my pr, the init_logging  func is invoked by process func, so the log before the process func will not write to the process.log file, such as `log.info("Processing analysis data")`, `log.info("Processing analysis data for Task #%d", task.id)`

doomedraven:
thanks for more info, will check again in a bit

qux-bbb:
> thanks for more info, will check again in a bit

If there is any question, please ask me.

doomedraven:
there are no questions, there is lack of time so i must manage priorities, and this one must be tested, don-t forget we doing this in our free time, we have our jobs who pays our salaries, i know is small PR, but there some stuff that I must verify before, but sadly this weeks not that much time

qux-bbb:
> there are no questions, there is lack of time so i must manage priorities, and this one must be tested, don-t forget we doing this in our free time, we have our jobs who pays our salaries, i know is small PR, but there some stuff that I must verify before, but sadly this weeks not that much time

Got it. I also do this in my free time. So please relax and do it later.

doomedraven:
im relaxed lol, but as you push over and over i had to clarify you that is not our paid job. Trust me the latest thing that i want is to have PRs be on hold

doomedraven:
thanks, sorry for delay

PR #2037, opened 2024-04-01
Database and scheduler overhaul

tbeadle:
First off, I apologize for the size of this diff.

See the individual commits for a little more detail about the specific changes, but this PR aims to do a few things:

- Make maintenance of the scheduler simpler
- Make interaction with the database use the "unit of work" pattern so that transactions can be rolled back easily if an error occurs during some piece of it.
- Remove the need for the machine lock and "scheduling" of machines for tasks, while keeping it performant.
- Improve test coverage and make testing easier because configs and the database will start at known, clean state.

This must not be merged without vetting it in the staging environment for a while. I would love it if people using a cloud environment for the VM's could try it out as well because I do not have a way to test that.


modules/machinery/az.py
```diff
@@ -483,31 +482,6 @@ def availables(self, label=None, platform=None, tags=None, arch=None, include_re
             label=label, platform=platform, tags=tags, arch=arch, include_reserved=include_reserved, os_version=os_version
         )
 
-    def acquire(self, machine_id=None, platform=None, tags=None, arch=None, os_version=[], need_scheduled=False):
```
cccs-kevin, line 486: @cccs-mog note that this method is removed when testing

kevoreilly:
I am seeing:
```
  File "/opt/CAPEv2/lib/cuckoo/core/analysis_manager.py", line 497, in run
    self.launch_analysis()
  File "/opt/CAPEv2/lib/cuckoo/core/analysis_manager.py", line 459, in launch_analysis
    success = self.perform_analysis()
  File "/opt/CAPEv2/lib/cuckoo/core/analysis_manager.py", line 445, in perform_analysis
    self.run_analysis_on_guest()
  File "/opt/CAPEv2/lib/cuckoo/core/analysis_manager.py", line 402, in run_analysis_on_guest
    options = self.build_options()
  File "/opt/CAPEv2/lib/cuckoo/core/analysis_manager.py", line 236, in build_options
    "options": self.get_machine_specific_options(self.task.options),
AttributeError: 'AnalysisManager' object has no attribute 'get_machine_specific_options'
```
Can't see ``get_machine_specific_options()`` in the code

tbeadle:
I'll be pushing up another commit shortly.

PR #2041, opened 2024-04-03
Staging

doomedraven:


cccs-mog:
So far:
Scaling semaphore is not working with dynamic number of machines, basically the machine manager options as there is no machines parameter when using a autoscaling/dynamic number of machines. 

Getting a warning: `WARNING: Rooter returned error: inetsim_disable() missing 4 required positional arguments: 'inetsim_ip', 'dns_port', 'resultserver_port', and 'ports'`

I was not yet able to test the rest as for unknown reasons the database seem not even initialized correctly ie get stuck at __getattr__ for every method call, seem incomplete to me ? 

Am I missing something ?

doomedraven:
i looked a bit into inetsim, but didn't see what is wrong, once i will have some spare time again i will check that

tbeadle:
> So far: Scaling semaphore is not working with dynamic number of machines, basically the machine manager options as there is no machines parameter when using a autoscaling/dynamic number of machines.

Could you please explain a little more about this? I don't quite understand.

> I was not yet able to test the rest as for unknown reasons the database seem not even initialized correctly ie get stuck at **getattr** for every method call, seem incomplete to me should have the params of the function passed down as well in parenthesis like this getattr(_DATABASE, attr)(params) ? ?
> 
> Am I missing something ?

Do you have a traceback or any other information that you can give? `Database.__getattr__` should not be passing any other parameters. It is only proxying the getattr call to the _Database object. So if you do this:
```
db = Database()
db.some_func(param1)
```
accessing `db.some_func` will call the `__getattr__` function, which will get that method from the _Database class and then pass `param1` to it.

lib/cuckoo/common/abstracts.py
```diff
@@ -107,42 +106,44 @@ class Machinery:
...
 
-    def set_options(self, options: dict):
+    def set_options(self, options: dict) -> None:
         """Set machine manager options.
         @param options: machine manager options dict.
         """
         self.options = options
+        mmanager_opts = self.options.get(self.module_name)
+        if not isinstance(mmanager_opts["machines"], list):
+            mmanager_opts["machines"] = str(mmanager_opts["machines"]).strip().split(",")
```
cccs-mog, line 130: This is not with dynamic machinery like AZ and AWS, hence creating a problem with parsing this.

cccs-mog:
Yes you are right regarding the Database class and __getattr__, it seem my call were not processing correctly because of Rooter returned error: inetsim_disable() missing 4 required positional arguments: 'inetsim_ip', 'dns_port', 'resultserver_port', and 'ports' which is now fixed. I will continue testing but so far work great minus small details like the one flagged below !

cccs-mog:
More update: It seem that only one vmss(azure) is being initialized at a time not multiples like it should be. Available Specific Machines: {'win10x64': 10, 'windows': 10};. Can't seem to figure out why this is happening, will try to dig into this. Might be related to with the problem I am also facing with the number of pending tasks staying to 0 and very few tasks are being processes/active at a time. Let me know if you need some more information to investigate and if this is something specific to AZ/dynamic ? The database is full of tasks and only the specific vmss machines are listed there.

lib/cuckoo/common/abstracts.py
tbeadle: Would you mind trying out this patch?
[machine_lock.patch.txt](https://github.com/kevoreilly/CAPEv2/files/14909811/machine_lock.patch.txt)

cccs-mog: Seem to initialize fine, thanks for patching !

cccs-mog: DEBUG: # Active analysis: 1; # Pending Tasks: 0; # Specific Pending Tasks: {}; # Available Machines: 10; # Available Specific Machines: {'win10x64': 9, 'windows': 9, 'ub20x64': 1, 'linux': 1}; # Locked Machines: 1; # Specific Locked Machines: {'win10x64': 1, 'windows': 1}; # Total Machines: 11; # Lock value: 2/2 It seem the lock value doesn't follow correctly the number of available machines. I guess this is a work of progress because of this : ```
 #Here be dragons! Making these calls on the ScalingBoundedSemaphore is not
                # thread safe.
```

Comment (tbeadle) [reply lib/cuckoo/common/abstracts.py:130]: Could you try adding some debug logging in `update_limit()` and `check_for_starvation()` to make sure that they are getting called every 10 seconds (or whatever you have `scaling_semaphore_update_timer` set to) and make sure they're getting called with accurate parameter values?

Comment (cccs-mog) [reply lib/cuckoo/common/abstracts.py:130]: So the thread in question which invoke `update_limit() `and `check_for_starvation()` is just never started (missing a start()), so this is the source of this problem !

Comment (tbeadle) [reply lib/cuckoo/common/abstracts.py:130]: I've created https://github.com/kevoreilly/CAPEv2/pull/2055 to start the thread.

**Comment by tbeadle** [comment]
@cccs-mog An update to staging has been applied to start the thread. Could you test it out again? I appreciate your patience on this!

**Comment by cccs-mog** [comment]
Observation based on the last version: 
-Restarting cape is not cleaning up correctly, the process is not fully killed so the port is still in used.
-Restarting cape on AZ give an error because I think it doesn't close correctly the vmss(s).
-Still having issue with submitting tasks

Regarding the thread and the semaphore, it seem fine now thanks for fixing it ! @tbeadle

**Comment by tbeadle** [comment]
@cccs-mog 
> -Restarting cape is not cleaning up correctly, the process is not fully killed so the port is still in used.

Which port are you referring to? How are you attempting to stop cape?

> -Restarting cape on AZ give an error because I think it doesn't close correctly the vmss(s).

Can you give more details about the error you're seeing (e.g. a traceback). I don't use azure at all. I've looked at the code again and everything *seems* sane, but I must be missing something.

> -Still having issue with submitting tasks

Earlier you said the database is full of tasks, but the pending count is 0. What state are those tasks in?

What exactly happens when you try to use submit.py?

Thanks! Unfortunately, I won't be able to respond until after Sunday, though.

**Comment by doomedraven** [comment]
hello guys, not sure how this is relevant to this branch, but i see that on master it doesn-t work neither, i had pending ´static´ tasks that wasn't processed and it wasnt going to task that need vm, looking into code what changed to try to spot where we changed that

**Comment by cccs-mog** [comment]
1) I am talking about port 2042 and I am killing the service via systemctl. From the process viewer it seem that some threads are still running. 
2)Regarding azure the only trace or hint of the problem is` [root] CRITICAL: CuckooCriticalError: No machines available `, which is solved by creating a new vmsss for Azure, note that I am using reset_pool_size=true. This behavior was not there consistently before trying this PR so this is why this is mentionned. I am trying to debug the issue but might take some time. 
```
File "CAPEv2/cuckoo.py", line 143, in <module>
    cuckoo_main(max_analysis_count=args.max_analysis_count)
  File "CAPEv2/cuckoo.py", line 104, in cuckoo_main
    sched.start()
  File "CAPEv2/lib/cuckoo/core/scheduler.py", line 289, in start
    self.machinery_manager.initialize_machinery()
  File "CAPEv2/lib/cuckoo/core/machinery_manager.py", line 241, in initialize_machinery
    raise CuckooCriticalError("No machines available")
lib.cuckoo.common.exceptions.CuckooCriticalError: No machines available
```
3)The status is "reported" which I think is a new status ?

doomedraven:
i see a logic issue in general here and in master, about pending any that is not file category, like static, pcap
If this code is not satisfied `relevant_machine_is_available = self.db.is_relevant_machine_available(task)` it won't process task, but if task is static, pcap it wont get futher to another tasks, im trying to spot where code changed, i probably will add quick check of category. There is `category_checks` but is too late, and never really reaches to it

`if self.task.category in ("file", "pcap", "static"):`

tbeadle:
I saw that particular issue while I was working on the scheduler and think
that it should be addressed in staging.

doomedraven:
but i would love if we could think about optimize case where we have static jobs, all VMs are running thats fine, but static jobs are not processed, and while tasks are in VM, servers are idling when they could process, but we maybe can do this once the main feature is done

tbeadle:
> but i would love if we could think about optimize case where we have static jobs, all VMs are running thats fine, but static jobs are not processed, and while tasks are in VM, servers are idling when they could process, but we maybe can do this once the main feature is done

This should already be getting taken care of in the code on staging. See https://github.com/kevoreilly/CAPEv2/blob/cc4d767c96cf682a166f0ed7a69aed770959f303/lib/cuckoo/core/scheduler.py#L179

tbeadle:
> 1. I am talking about port 2042 and I am killing the service via systemctl. From the process viewer it seem that some threads are still running.

Could it be that you have tasks running when you attempt to stop the service and just aren't waiting long enough? A normal SIGTERM, as issued by `systemctl stop` will allow the tasks that are currently running to finish and then the process will exit. See https://github.com/kevoreilly/CAPEv2/blob/cc4d767c96cf682a166f0ed7a69aed770959f303/lib/cuckoo/core/scheduler.py#L270. 

>    2)Regarding azure the only trace or hint of the problem is`[root] CRITICAL: CuckooCriticalError: No machines available`, which is solved by creating a new vmsss for Azure, note that I am using reset_pool_size=true. This behavior was not there consistently before trying this PR so this is why this is mentionned. I am trying to debug the issue but might take some time.

Again, I've never used Azure, but it would seem to me like having a vmss available for Cape to use (if I'm correctly understanding what a vmss is) is a reasonable requirement, so this may be a "works as expected" type of thing. I'm not sure.

> 3)The status is "reported" which I think is a new status ? The tasks get added to the DB, I get the tasks successfully created but the queue is still 0. The tasks in the DB are in the "reported" state.

"reported" is not a new status. It is what the task goes to after cape-processor is finished with it. Out of curiosity, are these tasks submitted for static analysis, as opposed to dynamic analysis that requires a VM?

doomedraven:
@cccs-mog was you able to check this? It would be nice to merge with but ensuring that it works for you too

cccs-mog:
Hi @doomedraven and @tbeadle, I tested the new version and I am still facing the same issues. Haven't had much time to look into it yet, the issue seem to be the status not getting reflected correctly for tasks and machines. The status for new task is "reported" like mentionned before and they are submitted for dynamic analysis. Regarding the machines the status is "f" which is obviously not a valid status. Not sure how the machinery would be affecting the tasks status. Looking into it.

nbargnesi:
> Regarding tasks: Now I am totally unable to create tasks using submit utilities

I see this as well. I think we just need to wrap a transaction around the db calls in `utils/submit.py`:
1. [add_url](https://github.com/kevoreilly/CAPEv2/blob/791b0609a300b61b7ae3b52d5c3c2007597033ed/utils/submit.py#L208)
2. [demux_sample_and_add_to_db](https://github.com/kevoreilly/CAPEv2/blob/791b0609a300b61b7ae3b52d5c3c2007597033ed/utils/submit.py#L342)

cccs-mog:
Thanks for fixing this. 
Now while running tasks I get this annoying messages:
2024-05-08 21:03:55,649 [lib.cuckoo.core.machinery_manager] DEBUG: Task #XX: no machine available yet for task requiring machine '', platform '' or tags '[<Tag(2,'win10x64')>]'. It doesn't seem to affect the analysis. Is this an expected behavior with this new way of scheduling ? If so maybe this error message should be limited/disabled ?

doomedraven:
ouch is 2w since last commit, any progress here guys? i would love to merge it but dont want to break anyones sandbox

doomedraven:
@cccs-mog im gonna merge this to master but please be on hold for 1-2w or test it on dev, we gonna help you to get it working on cloud, but i need this in as there a lot of good code

cccs-mog:
I am running with debug yes, so if there is no particular issue on that front this is fine with me. Only remaining issue is the machines not getting created on restart unless I switch vmss, its between the initialization and the prompt for machines, it get emptied out which make it fail.I think it's really the deletion which is happening between the machinery check. I will try assuming it's empty on az side for every restart to see if this resolve the issue. Otherwise it seem usable and will continue using it on dev for a few week and reach out for any issue. Thanks everyone !

PR #2050, opened 2024-04-09
Make the guest.platform column non-nullable.

tbeadle:
The original migration to create this column set it to nullable until default values could be populated, but since it's not nullable as defined in database.py, the migration should set it back to non-nullable after filling the existing rows with the default value.


tbeadle:
I just realized that this will not work on sqlite. It does not support `alter column`.

PR #2054, opened 2024-04-10
Fill categories when fetching signature data for scoring

ibaiape:
The current scoring script relies on signature categories, which currently are not filled when getting the signature info through `as_result()`. This fixes the score generation for executables on reports, and should not break if any signature doesn't define categories as it would simply provide an empty list.

It might be worth adding any of the other signature fields which aren't currently filled, but as I don't know if this should be the case and I currently only needed the malscore I haven't checked for the rest.


doomedraven:
thanks

PR #2056, opened 2024-04-11
fix a few Windows analyzer bugs, add test coverage

nbargnesi:
Details are in the commit, but at a high-level this does the following:

- starts to break up the huge `Analyzer.run()` method into small, testable chunks
- fixed a handful of small bugs
- adds lots of basic tests to ensure we don't break things in the future as we improve this even more


doomedraven:
ready to merge right?

nbargnesi:
yes, ready to go

doomedraven:
thank you for this

PR #2057, opened 2024-04-11
Add docker stuff

nbdy:
CAPE can run in a docker container.
With the docker-compose.yml you can run the web-interface, result server and database with one command (`docker compose up`).
You should still adjust the configuration files in the 'docker/conf' directory which is mounted when using docker compose.

The only parameter that has changed from the defaults is the database connection string.
Also I've removed the comments from the kvm.conf file so the block can be copy-pasted easily.

It sure would be nice to have a settings interface in the web ui at some point.
I'm looking forward to using this with Unraid.


doomedraven:
Hello, thank you, this is nice add for dev server, so i will have add note, as it not exposes cape for "production" with nginx and other to handle load properly. Also kinda interesting is how it handles load of many vms and tcpdumps

im a bit busy those days but i will hope to merge this next week with small changes, did you see also this one? https://github.com/CAPESandbox/CAHI ?

about configs, instead of create tons of new config, if you want im fine to add block to `kvm.conf.default` with no comments for easier c&p

nbdy:
Sorry, I did not and probably won't have any time for this in the future since I've shifted my focus to other projects. I'll close this PR since it has nearly been a year now.

PR #2059, opened 2024-04-12
fix: allow partial api name, search button does not work, allow space between names

jonisworking:
1. Allow partial API names:
> NtDelayExecution can now be searched using "NtDel" instead. This also resolved the case-sensitive issue ntdelayexecution = NtDelayExecution. 

2. Search button does not work:
> Clicking the physical button did nothing

3. Allow space between names:
> In case of unintentional spaces in between api names, originally "xx,xx" and now allows for "xx, xx" as well


doomedraven:
thank you, we had this in todo for so long, but never had time to put prior on this.

kevoreilly:
Thank you 🙏

PR #2060, opened 2024-04-12
feat: ranged task delete for api

jonisworking:
Allow deleting jobs from the API using ranges (easier), and update the documentation to reflect the current option/requirement


doomedraven:
thanks

kevoreilly:
👌

PR #2074, opened 2024-04-19
Update disable firewall command

qux-bbb:
This link says that we should use advfirewall rather than firewall:  
https://learn.microsoft.com/en-us/troubleshoot/windows-server/networking/netsh-advfirewall-firewall-control-firewall-behavior


doomedraven:
tnx

PR #2078, opened 2024-04-22
Refactor Linux strace

winson0123:
Previous implementation outputted strace logs onto the Linux guest VM. This refactor allows the strace logs to streamed directly to the CAPE result server on the host while the sample is running.

### DIFF
--- a/analyzer/linux/lib/common/results.py
+++ b/analyzer/linux/lib/common/results.py
@@ -58,9 +58,8 @@ def append_buffer_to_host(buffer, nc=None):
         while idx < size:
             nc.send(buffer[idx : idx + BUFSIZE], retry=True)
             idx += BUFSIZE
-    except Exception as e:
-        log.exception("Exception uploading buffer %s to host.")
-        log.exception(e)
+    except Exception:
+        log.exception("exception uploading buffer to host")
 
     nc.buffer_size += size
 
--- a/analyzer/linux/lib/common/results.py
+++ b/analyzer/linux/lib/common/results.py
@@ -53,13 +53,13 @@ def append_buffer_to_host(buffer, nc=None):
         log.warning("Buffer is too big: %d; max size: %d", size + nc.buffer_size, config.upload_max_size)
         return
 
-    try:
-        idx = 0
-        while idx < size:
-            nc.send(buffer[idx : idx + BUFSIZE], retry=True)
+    idx = 0
+    while idx < size:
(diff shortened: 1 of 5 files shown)


kenleejl:
Great Job!!!! Thanks @winson0123 !

BaconPotatoCat:
Wow!!! Amazing work! Thank you @winson0123

analyzer/linux/lib/common/results.py
```diff
@@ -44,13 +44,41 @@ def upload_to_host(file_path, dump_path, pids="", ppids="", metadata="", categor
...
+    if config.upload_max_size < size + nc.buffer_size and not config.do_upload_max_size:
+        log.warning("Buffer is too big: %d; max size: %d", size + nc.buffer_size, config.upload_max_size)
+        return
+
+    try:
+        idx = 0
+        while idx < size:
+            nc.send(buffer[idx : idx + BUFSIZE], retry=True)
+            idx += BUFSIZE
+    except Exception as e:
```
nbargnesi, line 61: Might be some room for improvement here - couple suggestions:

- Move the try/except around `nc.send()` where you're expecting exceptions
- If `nc.send()` does fail, I doubt `retry=True` is helpful and only a piece of the buffer isn't too useful. Maybe make this an error and return from the function?

analyzer/linux/lib/core/packages.py
```diff
@@ -118,8 +119,11 @@ def start(self):
             # Remove the trailing slash (if any)
             self.target = filepath.rstrip("/")
         self.prepare()
-        # self.normal_analysis()
         self.strace_analysis()
+        self.nc.init("logs/strace.log", False)
+        self.thread = Thread(target=self.thread_send_strace_buffer)
+        self.thread.start()
```
nbargnesi, line 125: This thread will start immediately and potetenially miss `self.proc.stderr` when it runs. I think you want to synchroinize its execution with `strace_analysis`. Have a look at [threading event objects](https://docs.python.org/3/library/threading.html#event-objects). You can hold back the thread's execution until self.proc is readable.

And with self.proc.stderr being line-buffered I think - you can probably set a threading event each time a line is ready for reading.

doomedraven:
this is fine to go right?

winson0123:
Unless there's an issue with not passing the tests, I think it is good to go

doomedraven:
thank you

PR #2079, opened 2024-04-22
Add MongoDB indices for analysis fields

JSCU-CNI:
This PR adds `keyword` indices for certain fields in the `analysis` collection. This massively improves load time when accessing an individual analysis result in CAPE on large MongoDB instances. 

CAPE uses one large OR-query when loading an analysis result which is why we opted for a separate index for every field on which the aggregation is run. The query also looks for exact matches so the fields are indexed as `keyword` instead of `text`.

### DIFF
--- a/lib/cuckoo/core/startup.py
+++ b/lib/cuckoo/core/startup.py
@@ -92,20 +92,40 @@ def check_webgui_mongo():
         client = connect_to_mongo()
         if not client:
             sys.exit(
-                "You have enabled webgui but mongo isn't working, see mongodb manual for correct installation and configuration\nrun `systemctl status mongodb` for more info"
+                "You have enabled webgui but mongo isn't working, see mongodb "
+                "manual for correct installation and configuration\n"
+                "run `systemctl status mongodb` for more info"
             )
 
-        # Create an index based on the info.id dict key. Increases overall scalability
-        # with large amounts of data.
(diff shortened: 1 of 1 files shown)


lib/cuckoo/core/startup.py
```diff
@@ -92,20 +92,40 @@ def check_webgui_mongo():
...
+            "procdump.extracted_files.sha256",
+            "procmemory.extracted_files.sha256",
+            "target.file.file_ref",
+            "dropped.file_ref",
+            "CAPE.payloads.file_ref",
+            "procdump.file_ref",
+            "procmemory.file_ref",
+        ]
+        for item in items:
+            mongo_create_index(
```
nbargnesi, line 123: This will take a _very_ long time for a lot of CAPE installations - terabyte scale is common. Probably better to do this in the utils somewhere and emit a warning if the indices aren't found.

nbargnesi: I'd suggest as a new module in [utils](https://github.com/kevoreilly/CAPEv2/tree/master/utils), `mongodb_indices` or something appropriately named.

In the startup module you can check the indices are there and emit a warning if they're missing.

doomedraven:
this is obsolete now, i have added indexes for md5 and sha256 + rewrote the searching

PR #2082, opened 2024-04-23
Added more tests for the analyzer

rkoumis:
- fixed a couple of bugs
- had a couple of questions

Bugs:
- `_handle_ksubvert()` was iterating over the process list while modifying it, which usually does not work
- in `_handle_shutdown()` we were trying to `dump_files` on the CommandPipeHandler instead of the analyzer
- in `_inject_process()` we weren't passing the thread_id properly to the new Process object
- in `_handle_process()` we weren't properly updating the analyzer's `LASTINJECT_TIME`

Questions:
- in `_handle_interop()` should we check if `ANALYSIS_TIMED_OUT` before we do anything?
- in `_inject_process()` do we need to update `LASTINJECT_TIME` ?

FYI @nbargnesi @tbeadle @josh-feather

### DIFF
--- a/analyzer/windows/analyzer.py
+++ b/analyzer/windows/analyzer.py
@@ -212,11 +212,14 @@ class Analyzer:
         self.do_run = True
         self.time_counter = 0
 
+        self.command_pipe = None
+        self.default_dll = None
         self.process_lock = Lock()
         self.files_list_lock = Lock()
         self.pid = os.getpid()
         self.ppid = Process(pid=self.pid).get_parent_pid()
         self.files = Files()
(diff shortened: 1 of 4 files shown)


kevoreilly:
There is no point in handling interop if the analysis has timed out so 1) yes

2) yes

rkoumis:
Work in progress: https://github.com/rkoumis/CAPEv2/pull/15 - It should be ready for you tomorrow.

rkoumis:
Instead of this PR, I opened https://github.com/kevoreilly/CAPEv2/pull/2083 against staging.

PR #2090, opened 2024-04-28
Zloader update

enzok:


kevoreilly:
This PR has broken all cape monitor sigs since it was merged as the monitor yara does not even compile!

```
error: rule "Zloader_2024" in D:\work\CAPE\CAPE\master\CAPEv2\analyzer\windows\data\yara\Zloader.yar(25): undefined string "$decrypt_conf*"
```

PR #2111, opened 2024-05-08
Fix submit.py

tbeadle:
Update submit.py so that it properly uses database transactions to commit the changes. Fixes the issue in https://github.com/kevoreilly/CAPEv2/pull/2041#issuecomment-2101251635

Pinging @cccs-mog for awareness.


cccs-mog:
Thanks ! Will be testing with that.

PR #2116, opened 2024-05-10
Feature to ability monitor traffic between cape result server and sandbox.

piolug93:


doomedraven:
fix conflicts plz, and do not modify `conf/routing.conf` those config shouldnt be touched, only .default

doomedraven:
and what is adventage of using nat instead of hostonly?

piolug93:
Okay will resolve conflicts. Hostonly is still used, while MASQUERADE is not used in iptables. Another feature is the use of vrf so that traffic to resultserver passes by default gateway instead of through the interface lo.

nbargnesi:
Great idea @piolug93. There's a lot of interesting stuff going on behind the scenes - alongside the result server traffic, agent communication and layer 2 traffic would be helpful to expose in analyses.

Did you consider adding an auxiliary module alongside the [existing sniffer](https://github.com/kevoreilly/CAPEv2/tree/master/modules/auxiliary)?

The existing sniffer [heavily filters traffic](https://github.com/kevoreilly/CAPEv2/blob/9c8d6da44b595f8140a5cd76edd8101f6812c3b0/modules/auxiliary/sniffer.py#L111) (also [#L139](https://github.com/kevoreilly/CAPEv2/blob/9c8d6da44b595f8140a5cd76edd8101f6812c3b0/modules/auxiliary/sniffer.py#L139), [auxiliary.conf](https://github.com/kevoreilly/CAPEv2/blob/9c8d6da44b595f8140a5cd76edd8101f6812c3b0/conf/default/auxiliary.conf.default#L61)) - having the ability to monitor result server, agent, and layer 2 traffic with auxiliary module and write out an analysis-debug.pcap to capture it all would be fantastic!

piolug93:
Good to know about it, I had no idea about it. I haven't plans for write auxiliary module because i no need  see that traffic in CAPE.

lib/cuckoo/core/startup.py
```diff
@@ -424,12 +425,16 @@ def init_routing():
...
+        if routing.routing.nat:
+            # Disable & enable NAT on this network interface. Disable it just
+            # in case we still had the same rule from a previous run.
+            rooter("disable_nat", routing.routing.internet)
+            rooter("enable_nat", routing.routing.internet)
+            # Populate routing table with entries from main routing table.
+        else:
+            rooter("disable_nat", routing.routing.internet)
+        if routing.routing.no_local_routing:
+            rooter("init_vrf", routing.routing.rt_table, routing.routing.internet)
```
xiangchen96, line 437: Hi @piolug93 
now that is on master we are having some conectivity issues, shouldn't no_local_routing logic be inside the else?
```
else:
    rooter("disable_nat", routing.routing.internet)
    if routing.routing.no_local_routing:
        rooter("init_vrf", routing.routing.rt_table, routing.routing.internet)
```

piolug93: What problem do you have ? What you have set in routing.conf for nat and no_local_routing ?

piolug93: It's good observation. That should be in else block in accordance with comment from routing.conf for no_local_routing value.

piolug93: @doomedraven  thx for fix. 
https://github.com/kevoreilly/CAPEv2/commit/925ba96aba58057b1d1bae119063fdf41bc0c506

doomedraven: Thank you both

PR #2121, opened 2024-05-15
added nginx config enable in docs

Zer0Risk:
I added 2 bash lines in the web Documentation, so that users know they also have to enable the newly created nginx config.  This is mandatory in order for the nginx config to work.

### DIFF
--- a/docs/book/src/usage/web.rst
+++ b/docs/book/src/usage/web.rst
@@ -206,6 +206,14 @@ Replace ``www.capesandbox.com`` with your actual hostname.
         }
     }
 
+Now enable the nginx configuration by executing the following: 
+
+.. code:: bash
+
+   rm -f /etc/nginx/sites-enabled/default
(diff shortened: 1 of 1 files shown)


doomedraven:
thanks, but i guess in this case better is to overwrite default from install script

Zer0Risk:
Thanks for the feedback. Since Nginx is only needed for optional interactive sessions, modifying the main installation script isn't ideal. The documentation helps users configure Nginx only if they choose this feature, ensuring clarity and flexibility.

z00mz00m95:
Hello everyone, I can only confirm this PR. The documentation for the interactive sessions describes that you have to create a configuration file at /etc/nginx/sites-available/cape but not that it has to be activated. For now, the nginx process is not automated, so this part should be included in the documentation to make this clear.

doomedraven:
thanks for PR

PR #2134, opened 2024-05-23
feat(linux): add linux packages to submission form

xiangchen96:


web/submission/views.py
```diff
@@ -57,17 +57,22 @@
 logger = logging.getLogger(__name__)
 
 
-def get_form_data(platform):
```
xiangchen96, line 60: removed platform arg, as we were already listing all machines anyway

doomedraven: thanks for clarification

PR #2141, opened 2024-05-27
Bug fix in qemu shutdown

xalicex:
- Modification of the "stop" function to patch unstoppable VM bug.
- Adding "-monitor stdio" argument  to be able to send "qemu monitor" commands


doomedraven:
thanks

PR #2142, opened 2024-05-28
Scheduler fix

tbeadle:
The previous change to find_next_serviceable_task has a bug if
allow_static is True. In this case, we would only call
find_pending_task_not_requiring_machinery and never call
find_pending_task_to_service.
Instead, we want to track if the maximum number of machines has been
reached and use that information to make the determination of which
function to call.

The calls to main_db.set_status in node_submit_task should not be
in a begin() block because node_submit_task is already called from
within a transaction.


doomedraven:
thank you

kevoreilly:
Thanks Tommy

PR #2143, opened 2024-05-29
behavior tree rewrite

doomedraven:


doomedraven:
To prevent recursive errors on json reports, for some malware spawns a lot of processes, and we embed children into parent process

xiangchen96:
This works, but changing the schema of the json has some implications:
- Will probably break things for users
- Dynamic CAPA will break: https://github.com/mandiant/capa/blob/master/capa/features/extractors/cape/models.py#L312-L319

I would prefer a conf that limits the depth of the tree with a sensible default or being able to disable it (users can rebuild the tree themselves from processes if needed).

doomedraven:
well sometime progress requires breaking changes, this rarely happens, but would be nice to handle this properly without loosing details, but yes quick fix i guess will be configurable limit, but that limit should be the same as json limit, so is only on/off as is uniq field that breaks

PR #2151, opened 2024-06-03
Fix enforce_timeout

xiangchen96:
`enforce_timeout` logic was removed here:

https://github.com/kevoreilly/CAPEv2/pull/2070/files#diff-5b414b99a827dfb21077aedddb85b188d90b6aeb8f4889bf06262ff2a7f44bbeR631-L608

Bringing it back


doomedraven:
thanks

enzok:
self.pid_check is initialized as false, is this doing anything?

xiangchen96:
in our case is was set to true in

```
if self.options.get("manual", False):
    self.pid_check = True
```

@enzok

PR #2175, opened 2024-06-17
Update cape2.sh

demon-i386:
added plocate (command locate) and libvirt-dev (python libvirt requirement)


doomedraven:
hello, we don't need `libvirt-dev`, what os do you use that requires that library?

demon-i386:
> hello, we don't need `libvirt-dev`, what os do you use that requires that library?

ubuntu 22.04

cccs-kevin:
I needed the `libvirt-dev` package also

doomedraven:
ok i did check, if you install `libvirt-dev` it forces install of `libvirt0`, any of my servers doesn't have that library. so i won't poison install script

PR #2176, opened 2024-06-17
Update cape2.sh

demon-i386:
added ninja-build to dependencies, required for pyre2 build.

 error: subprocess-exited-with-error
  
  × Building wheel for pyre2 (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [58 lines of output]
      running bdist_wheel
      running build
      running build_ext
      CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.


doomedraven:
thanks, pyre2 is dead, and not supported in upcoming versions of python, so i guess is better directly remove it instead of add it extra dependencies

PR #2199, opened 2024-07-01
Add optional process.file_activities to link pid with files

xiangchen96:


ethhart:
I know capa isn't used in your implementation, but am aware of some of the linkages and see some activity from you guys on the capa repo, so wanted to make you aware that this change breaks the cape model defined in the current version of capa https://github.com/mandiant/capa/blob/master/capa/features/extractors/cape/models.py#L301

Either the results shouldn't contain the empty dict or the capa model needs to be updated. I apologize that I am unable to open a PR.

PR #2200, opened 2024-07-01
Improve sample architecture detection

josh-feather:
- Samples are no longer automatically tagged with the x86 tag if they are not architecture specific
- Adds darwin inference for Macho-O samples; preventing Mach-O samples from being detonated on Windows VMs
- Consolidates architecture prediction and platform determination code by moving it into the File object
- Fixes a bug where PE32/PE32+ samples are always assigned x86/x64 tags
- Adds tests to validate platform and architecture detection code
- Disables magic type emulation for any PE that was not compiled for x86 or x64


josh-feather:
Going to test this in our environment for a few days before marking as ready. Please feel free to provide feedback in the meantime.

doomedraven:
thank you

PR #2204, opened 2024-07-02
yara-x-metadata

doomedraven:


doomedraven:
yara-x >=0.4 supports metadata

PR #2206, opened 2024-07-03
Add an option to fetch machine VNC port in machine view API

daniel-oronsi:
This PR adds the virtual machine's VNC port to the machine view API route, only when the corresponding flag is enabled in the route's configuration.

This allows users to dynamically connect to a virtual machine by querying the machine view API for the VNC port and using the provided port for the connection.

### DIFF
--- a/conf/default/api.conf.default
+++ b/conf/default/api.conf.default
@@ -271,6 +271,7 @@ rps = 1/s
 enabled = no
 auth_only = no
 rps = 1/s
+fetch_vnc_port = no
 #rpm = 10/m
 
 # Pull information about the Cuckoo host server.
--- a/web/apiv2/views.py
+++ b/web/apiv2/views.py
@@ -11,6 +11,7 @@ from datetime import datetime, timedelta
 from io import BytesIO
 from urllib.parse import quote
 from wsgiref.util import FileWrapper
+from xml.etree import ElementTree as ET
 
 import pyzipper
 import requests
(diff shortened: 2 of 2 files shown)


doomedraven:
hello, thank you, can you allow edit to maintainers? `fetch_vnc_port = yes` need to uncomment this and set to no + some other changes

daniel-oronsi:
Unfortunately, it looks like this option isn't available when the fork is inside an organization...
I’ve made the configuration changes you requested, are there any other changes you'd like me to make? :)

doomedraven:
the port is stored to options of the task, better would be to just get task details and parse options, i will change that once i have some spare time one of this days

https://github.com/kevoreilly/CAPEv2/blob/de34cf5aa6054104d149fb5319317f10fb30e1c4/lib/cuckoo/core/database.py#L741

doomedraven:
i still don't have permissions to edit your code

daniel-oronsi:
I can't add write permissions to maintainers since this fork is within our organization - it looks like this is a well-known issue:
https://github.com/orgs/community/discussions/5634

Would you rather I close this PR and you open a new one?

doomedraven:
i guess so, i will just add func to database get_vnc_port, similar to set_vnc_port as port is added to option so you don't need to go and check the vm configuration

doomedraven:
btw i see one issue here, why do you check the machine port using machine view? i mean, port can be changed per task, so is better to use `task_view` to get the port as it already exposed there. i just try to clarify

daniel-oronsi:
That's a good point. I initially thought it made sense for the VNC port to be part of the machine view since it's a property of the machine, but I agree that querying this data from the task itself makes a lot of sense...

Since this feature is relatively new, I didn't notice that the port is already available in the task view. I think I'll use the options from the task view to fetch the VNC port.

This PR is no longer necessary, so I'll close it. Thanks for your help!

doomedraven:
thats fine, well im open for new features, i just need to understand them :)

PR #2220, opened 2024-07-10
Added summary and description to packages

rkoumis:
- Every package now will have a summary
- Every package now will have a description
- Many packages now will have a list of option names


doomedraven:
thank you

PR #2234, opened 2024-07-16
Updates to detect-it-easy handling.

tbeadle:
Don't require the diec binary to exist since we're now using the python bindings.


doomedraven:
It doesn't work without diec installed right?at least when I tested it didn't work for me

PR #2255, opened 2024-07-31
Agnostic OS Qemu screenshots

dsecuma:
This PR adds a module to get screenshots from any OS that running in QEMU via libvirt. The idea to create this is based on the variety of desktops, X (Xorg,Wayland,...) and OS types and versions, to have one solution for all of them.

- The module uses some functionality from current windows screenshot auxiliary module to reduce the number of screenshots. In addition `deduplicity` also is executed in postprocessing. 

- `screenshots_linux` and `screenshots_windows` must be disabled.

Is open to changes, ideas and even considering moving to a community repo. The module has been tested on several Linux and Windows VMs.

Thank you.

### DIFF
--- a/conf/default/auxiliary.conf.default
+++ b/conf/default/auxiliary.conf.default
@@ -49,6 +49,11 @@ enabled = no
 # Enable or disable the use of an external sniffer (tcpdump) [yes/no].
 enabled = yes
 
+[QemuScreenshots]
+# Enable or disable the use of QEMU as screenshot capture [yes/no].
+# screenshots_linux and screenshots_windows must be disabled
+enabled = no
+
 # enable remote tcpdump support
 remote = no
 host = root@192.168.122.1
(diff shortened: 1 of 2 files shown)


nbargnesi:
This is similar to PR #1814 which has been around for nearly a year now: https://github.com/kevoreilly/CAPEv2/pull/1814

@doomedraven I'd suggest one approach, either what was authored in this one or #1814.

I'd suggest #1814 over #2255 since it's agnostic of the OS _and_ the machinery. We've got folks using Libvirt+QEMU, Azure, VirtualBox, and Hyper-V.

doomedraven:
@dsecuma did/do you have any problems with that screenshot feature that @nbargnesi mentioning?

dsecuma:
Oh yes, it's duplicated. Maybe I would just add a comparison to reduce the writes that the deduplicate module then deletes.

PR #2276, opened 2024-08-14
Disable Signatures Module if Benign File

para0x0dise:
Related to PR #1941, If the file is benign, I think we don't need to get any triggered signature to avoid false-positive detections.

### DIFF
--- a/lib/cuckoo/core/plugins.py
+++ b/lib/cuckoo/core/plugins.py
@@ -661,9 +661,14 @@ class RunSignatures:
 
         malscore, malstatus = calc_scoring(self.results, matched)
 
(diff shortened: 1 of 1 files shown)


kevoreilly:
Yeah malscore is off by default in cape! It's only included to keep a few folks happy, but it's rubbish compared to cape's classification methods. Signatures are a fundamental measure of behavior, taking this logic to the extreme we could delete all the api logs and payloads too based on a 'decision' by malscore. I don't think we would want to go down that road, rather keep cape's output honest based on measured behavior rather than some arbitrary notion of maliciousness or a poorly determined score.

PR #2283, opened 2024-08-21
Exclude the FOG Server IP

para0x0dise:


doomedraven:
well is kinda ok, but by default it will be `<FOG server IP>` so need some logic to check if that machinery is used first

Issue #2254, opened 2024-07-31
The AzSniffer module initialization failed, causing the initialization to return


JohnSmith054:
## About accounts on [capesandbox.com](https://capesandbox.com/)
* Issues isn't the way to ask for account activation. Ping capesandbox in [Twitter](https://twitter.com/capesandbox) with your username

## This is open source and you are getting __free__ support so be friendly!

# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [ ] I am running the latest version
- [ ] I did read the README!
- [ ] I checked the documentation and found no answer
- [ ] I checked to make sure that this issue has not already been filed
- [ ] I'm reporting the issue to the correct repository (for multi-repository projects)
- [ ] I have read and checked all configs (with all optional parts)

# Expected Behavior

Please describe the behavior you are expecting. __If your samples(x64) stuck in pending ensure that you set tags=x64 in hypervisor conf for x64 vms__

# Current Behavior

The AzSniffer module initialization failed with an error, causing the sniffer to not load properly, and as a result, the pcap file was not generated

# Failure Information (for bugs)

2024-07-31 20:05:18,651 [lib.cuckoo.core.plugins] ERROR: Failed to load the auxiliary module "<class 'modules.auxiliary.AzSniffer.AzSniffer'>": secret should be a Microsoft Entra application's client secret
Traceback (most recent call last):
  File "/opt/CAPEv2/lib/cuckoo/core/plugins.py", line 125, in start
    current = module()
  File "/opt/CAPEv2/modules/auxiliary/AzSniffer.py", line 48, in __init__
    self.credentials = self._get_credentials()
  File "/opt/CAPEv2/modules/auxiliary/AzSniffer.py", line 58, in _get_credentials
    return ClientSecretCredential(tenant_id=self.tenant_id, client_id=self.client_id, client_secret=self.client_secret)
  File "/home/cape/.cache/pypoetry/virtualenvs/capev2-t2x27zRb-py3.10/lib/python3.10/site-packages/azure/identity/_credentials/client_secret.py", line 47, in __init__
    raise ValueError("secret should be a Microsoft Entra application's client secret")
ValueError: secret should be a Microsoft Entra application's client secret

## Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. Using the default az.conf can trigger the bug.

#Sniffer configurations
# New sniffer logic implementing Azure Watchers packet capture with either local or blob storage file locations 
[...]

PR #2292, opened 2024-08-26
Add `enabled` checks in `init`, `start`, and `stop`.

ChrisThibodeaux:
Add `enabled` check on AzSniffer auxiliary module.

Relevant issue: https://github.com/kevoreilly/CAPEv2/issues/2254


doomedraven:
thanks, i guess the best will be to do the same that @enzok did yesterday to analyzer, to check if enabled before even import it

PR #2296, opened 2024-08-27
Support pcap_ng captures

enzok:


doomedraven:
oh that nice, but we don't write it right?

enzok:
No. Only temporarily.

josh-feather:
I've got some changes running locally that dump pcapng's with embedded TLS keys. I'll create an upstream PR at some point this week.

PR #2299, opened 2024-08-28
Fix AWS autoscaling premature machine deletion.

ChrisThibodeaux:
PR ready for review. AWS user confirmed changes fixed their premature machine deletion issue.


doomedraven:
thank you for update

PR #2301, opened 2024-08-31
Analysis Package for Batch Scripts

para0x0dise:


doomedraven:
hey, thank you, maybe instead of `bat.py` it should be called `batch.py`?

para0x0dise:
> hey, thank you, maybe instead of `bat.py` it should be called `batch.py`?

hmmm, I think you're right

doomedraven:
thanks

PR #2310, opened 2024-09-09
Update process.py for callback usage.

ChrisThibodeaux:


doomedraven:
i would ignore changing process.py, as callback and process sets the status, is not a big deal, but put specific community module into processing is no a good option

doomedraven:
the callback itself is updated already, thanks for PR to clarify what you mean for process.py but we can skip it

PR #2312, opened 2024-09-09
Add SslKeyLogFile aux module and PcapNg processing module

josh-feather:
Adds two new modules:
- SslKeyLogFile auxiliary module
- PcapNg processing module

SslKeyLogFile sets the SSLKEYLOGFILE environment var on the guest and collects the resulting log file. This is especially useful when detonating inside of browsers.

PcapNg takes SSL/TLS keys from tlsdump.log and SslKeyLogFile and injects them into the detonation PCAP using the `editcap` binary. The file is made available to download via the UI.


doomedraven:
thank you

PR #2317, opened 2024-09-14
fix: incorrect use of Config class.

bryanster:
The config class was used incorrectly to load the resultserver ip resulting in an error at db.add_machine because an empy Config object cannot be casted into the required type.

happend to ran into this issue and did some debugging to resolve the issue.


doomedraven:
Tnx

PR #2319, opened 2024-09-16
Prescript detection feature

cccs-mog:


doomedraven:
interesting feature, thank you

PR #2320, opened 2024-09-17
Fix CPUID replacement in kvm.c

YungBinary:
- Fix the path to kvm.c
- Fix the regex to match KVMKVMKVM\0\0\0


doomedraven:
thanks

Issue #2323, opened 2024-09-18
Database and AZ machinery issue 


cccs-mog:
# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [X ] I am running the latest version
- [X ] I did read the README!
- [X ] I checked the documentation and found no answer
- [X ] I checked to make sure that this issue has not already been filed
- [X ] I'm reporting the issue to the correct repository (for multi-repository projects)
- [ X] I have read and checked all configs (with all optional parts)

# Expected Behavior

A call to self.db.add_machine() should result on a machine being created and thus the check in initialize_machinery() listing machine should not induce an error for the AZ machinery.

# Current Behavior

No machine are created at all, the running of this function doesn't seem to produce any database commands being ran.

# Failure Information (for bugs)

The session goes all the way throught initialize_machinery(). The AZ machinery overwrite the initialize and initialize_check function. In all cases at some point the machinery will attempt to add machines to the database the same way as the other machinery would do it, but for some reasons it doesn't work. The connection to the database is ok because tasks can be added to it and the schema was created at the beginning. Tested on both SQLite3 and PosgreSQL. This could be a timing issue which is specific to scaling machinery but I would expect that explicitely calling the addition of a machine in the database to actually do it.

## Steps to Reproduce

1. Clean install of CAPE
2. Setup for the AZ machinery with VMSS
3. Run the cape service/daemon as normal

## Context
Fresh install on Ubuntu 20.04. Base installation made from cape2.sh. Every aspect seem to work fine except this. 

## Failure Logs

2024-09-18 18:26:45,412 [root] DEBUG: Imported "machinery" modules:
2024-09-18 18:26:45,412 [root] DEBUG:    `-- Azure
2024-09-18 18:26:45,413 [root] DEBUG: Checking for locked tasks...
2024-09-18 18:26:45,417 INFO sqlalchemy.engine.Engine BEGIN (implicit)
2024-09-18 18:26:45,417 [sqlalchemy.engine.Engine] INFO: BEGIN (implicit)
[...]

PR #2291, opened 2024-08-26
Azure standard vmss fix

cccs-mog:

PR #2324, opened 2024-09-21
Azure safety updates

ChrisThibodeaux:
Updates to handle critical errors when:
- Non-persistent machine deletion from the DB leads to a machine scheduled for deletion being assigned a task
- A machine's agent will sometimes be unreachable at startup
- A machine is sometimes added more than once to `delete_vm_list`
- On startup, an existing VM is on the VMSS
- On startup, a machine that was previously in the DB is still present on the VMSS
- On startup, when all initialized machines fail, add a configurable retry capability
- On startup, an existing VM is `locked`

Many of the changes are to make the code more thread-safe.
Delay monitor thread start to allow all VMSSs to be up and instances running, especially in the case of initial startup failures.


doomedraven:
Hello, thank you, i will need first move this to staging before it will be merged to master to ensure that we didn't break nothing

ChrisThibodeaux:
@doomedraven  Sounds good. I want to test it significantly on my end before moving forward with it as well. Is it possible to change this to a draft instead of a regular PR?

doomedraven:
yes then draft would be good

ChrisThibodeaux:
@cccs-mog If you have a chance to test this on your end, it should fix all the errors we discussed in #2323

ChrisThibodeaux:
@doomedraven I removed the changes in `scheduler.py` and `abstracts.py`. Only `machinery/az.py` has changes to it now.

Ready to move out of draft now.

doomedraven:
cool thank you, lets see if @cccs-mog can confirm it and then we fine to merge

doomedraven:
i changed from staging to master as changes are only to az.py, so they are not so dangerous for the rest of users and easier to review

cccs-mog:
Hi @doomedraven and @ChrisThibodeaux, I will take a look today and get back. Awesome work !

cccs-mog (commented):
Awesome ! Seem to work wonders ! Only missing subnet limits and quota from https://github.com/kevoreilly/CAPEv2/pull/2291 :) Thanks a lot !

doomedraven:
ok lets merge this then, and then someone use uses azure can add subnet limit and quota . thank you @ChrisThibodeaux for your work

PR #2329, opened 2024-09-25
CobaltStrike Beacon parser fix

enzok:


doomedraven:
test what is wrong with you it not fails in local :D

enzok:
looks like it failed in get_yara

PR #2332, opened 2024-09-27
Pending tasks timeout

cccs-mog:
Not sure how much this is needed but useful feature for cleaning the queue of old tasks while running automatically.


cccs-mog:
Missing the params in the test, will be adding them a bit later.

doomedraven:
let me know if is fine to merge

cccs-mog:
Oh thanks for fixing the test ! I did multiple tests and it seem to work fine. Might be good to have additional tests before merging. It should be good to go and shouldn't affect anything else however unless you activate the feature.

PR #2341, opened 2024-10-08
Change Creator ID and OEM Table ID in ACPI Tables

IridiumXOR:
Patch to hide QEMU from ACPI table IDs detection


doomedraven:
O thanks

PR #2356, opened 2024-10-15
Fix bug when download_file returns success.

tbeadle:


doomedraven:
ouch good catch, but we need to handle it, as zips with many files, could reject some of them. thanks to point to the problem, i will review it again tomorrow

doomedraven:
plz see my changes, now it should be properly handled and errors will be streamed down to user

PR #2368, opened 2024-10-23
YungBinary Add Stealc/KoiLoader Config Extractors

YungBinary:
Here's a preview of config extraction for StealC/KoiLoader :) been trying to find a job doing this full time so if anyone has any leads lmk!

Screenshot from 2024-10-23 09-38-52

PREVIEW CONFIG EXTRACTION


doomedraven:
Can you please add test for the extractor too?

YungBinary:
> Can you please add test for the extractor too?

On it, thanks!

YungBinary:
@doomedraven Okay I've added the tests, here's the PR for the test files repo: https://github.com/CAPESandbox/CAPE-TestFiles/pull/10

kevoreilly:
Thanks for this contribution 🙏 Unfortunately I am not keen on the idea of having decrypted strings in the config - for dynamically extracted strings there is already the facility to have them appear as a text file in the payloads tab. I am working with doomed on a potential mechanism to have strings dumped from static parsers to appear in a more suitable place than the config. For this reason I will hold off for now on merging this, and will be on leave for a week or so. But this addition is very much appreciated, I would request your patience while we work out the details.

YungBinary:
> Thanks for this contribution 🙏 Unfortunately I am not keen on the idea of having decrypted strings in the config - for dynamically extracted strings there is already the facility to have them appear as a text file in the payloads tab. I am working with doomed on a potential mechanism to have strings dumped from static parsers to appear in a more suitable place than the config. For this reason I will hold off for now on merging this, and will be on leave for a week or so. But this addition is very much appreciated, I would request your patience while we work out the details.

Thanks for the update. The goal was to have output useful for reverse engineers so they can pop the sample into IDA and be able to map labels like dword_x to specific strings, otherwise the strings aren't very meaningful on their own.

kevoreilly:
Yep absolutely - useful output no doubt. It's that the place for such output is really somewhere like the payload tab, just not the config.

YungBinary:
> Yep absolutely - useful output no doubt. It's that the place for such output is really somewhere like the payload tab, just not the config.

Okay gotcha that sounds good. I'll keep an eye out for DMs or messages here in the event y'all find something suitable and I need to make changes. Also, just wrote up a KoiLoader config extractor and test and added to this PR, can remove if it should be in its own PR? Thanks man!

doomedraven:
Hello, I would suggest to split commits per parser, as having both parsers here will be on hold till we have time to work on strings instead of directly merge the second one

doomedraven:
you don't need to fork repo twice, you can work on different branches

doomedraven:
@kevoreilly as quick workout for this i would just kick strings from web gui(allowing users to enable that in config if they want). so we can merge this and work on the move of strings properly once we have more spare time

doomedraven:
i have disabled string storing for now, we will be able to restore that later, so i can merge this PR :), thanks again @YungBinary

Issue #1716, opened 2023-08-25
[Feature Request] CAPE Extractor Output Standardization


cccs-rs:
Hey guys! Long time no see 🤓

# Expected Behavior

When running a CAPE extractor located in `./modules/processing/parsers/CAPE/`, we'd like to know what format to expect certain information in a deterministic manner. Similar to other extractor frameworks like MWCP and MACO.

# Current Behavior

Currently there's no way to tell the output format of the extractor without examining the source code which is inconvenient if trying to run multiple extractors in an automated system like Assemblyline.

# Possible Solutions?
 - CAPE creates its own output standard for it's extractors that are documented
 - CAPE could adopt an existing standard (like MACO) similar to the PR created a year ago, but this would imply there are also UI changes to account for the data in new format if used in `capesandbox.com`
 - The extractors could implement a secondary function that we can call manually to translate the configuration into another structured format 

## Context

Old PR: https://github.com/kevoreilly/CAPEv2/pull/1037

I look forward to picking up on our discussion!

doomedraven:
leaving this to @kevoreilly  as i don't use/cape about public extractors :)

cccs-rs:
@kevoreilly any spare time to discuss this??

cccs-rs:
That sounds great! Is there anything I can do on my side to help further along the implementation? Maybe add this new entrypoint to the existing parsers that you have?

doomedraven:
someone need to review all the parsers and make standard naming of things like: cnc, cncs, c2, c2s,  and being strings and lists etc. that would be first step, second step is easy wrapper that transforms that

cccs-rs:
As an aside, not super important, what do you guys think about putting the extractors in a separate repository and submoduling it into this repository. Similar to what you guys do with the test data?

Only reason why I ask is because we currently have to clone the whole repository just to get the extractors, which is a bit much but isn't the end of world if we have to.

cccs-rs:
> someone need to review all the parsers and make standard naming of things like: cnc, cncs, c2, c2s, and being strings and lists etc. that would be first step, second step is easy wrapper that transforms that

We can start doing a review on our side and try to start creating a standard for naming for things. Once we have that settled, we'll reach out to you guys to see what you guys think.

If everyone is happy with the naming, then we can start implementing the MACO conversion entrypoint in each parser and PR it. 

How does that sound?

doomedraven:
About review sound good, about side repo I don't care, that's Kevin's bit

[...]

PR #2373, opened 2024-10-27
Create MACO extractors that wrap the CAPE extractor

cccs-rs:
Related to: https://github.com/kevoreilly/CAPEv2/issues/1716

This should allow for converting the raw CAPE output to the MACO standard without interfering with the original extractor code.


doomedraven:
thank you, i guess it would be better to put this under community, as it almost 100 files. what do you think? i can merge and move

cccs-rs:
Could but the only problem is how to resolve the dependencies/imports if they're split across different projects.

Any thoughts on how I may do that? Is it possible to install CAPEv2 as a Python package perhaps?

doomedraven:
they will be pulled to the actual folder, it just to try to make more cleaner what is maintained by who, once they pulled, they will have all the access to the imported extractors

cccs-rs:
Totally makes sense, just thinking about how this is going to be used with the extractor CLIs (maco, configextractor-py) that are in projects like Assemblyline where the installation of the extractor is all around Python PATH manipulation and module loading (and in some cases, analysts don't want to install all of CAPE just to get access to the extractors).

Currently the process is (essentially what these [tests](https://github.com/CybercentreCanada/configextractor-py/blob/823ec6ae4a6a088b6d86a325dc6c692361199931/tests/test_detection.py#L133) do):

- Git clone / HTTP GET extractor repository
- Use CLI to point to directory containing extractors/module context
  - CLI looks for requirements.txt or pyproject.toml to determine package dependencies for extractors to install in a venv
  - CLI loads/keeps modules that match the extractor framework it's looking for

Also now that I think about it, automatic deployment in air-gapped systems would also pose an issue but there could be ways to get around that (ie. custom steps specifically for the CAPEv2 project but I'd prefer to stay as general as possible). 🤔

cccs-rs:
I'll see if I can think of something as this is mostly a problem with automatic integration/patching with our systems. 🤓

doomedraven:
yep, lets thing of that in one of the next day, as my mind is occupied by another subject today, so we can find a proper solution

kevoreilly:
Thanks for the PR - I really like this implementation, nice wrappers that people can use or not as they please, leaving the core nice and clean ❤️

cccs-rs:
As a point of inquiry, is there any negatives of keeping the extractors within the main project?

doomedraven:
no, the idea is just to properly try to split things,but we probably will have to merge and keep it here

doomedraven:
i have a question, do you use this script? https://github.com/CAPESandbox/community/blob/master/utils/download_parsers.py

doomedraven:
so if i understand properly

1. having that extractor fetcher we could add maco folder into there from community repo so it would get you all parsers. 
2. In air gap setup you still need to clone main repo, so you can clone both of them
3. It uses maco dependency, which probably only will be used by CCCS so i don't see reason to put it into core for everyone, i personally use `community_blacklist.py` for folders/files that i don't want on my servers
4. another thing that would be nice for you i guess, is splitted dependencies, to have extractor dependencies independent to main pyprojects so that would make smaller main pyproject, but main dependency i guess is yara y pycryptodomex for most of the cases. but you could have pure extractor related dependencies
5. if you prefer we can talk about this in pm

cccs-rs:
> 3\. It uses maco dependency, which probably only will be used by CCCS so i don't see reason to put it into core for everyone, i personally use `community_blacklist.py` for folders/files that i don't want on my servers

Well in this case, we could make `maco` an optional dependency similar to what was done from RKP so that way it doesn't impact CAPE and we could just install those dependencies in addition to what's part of the project by default.

What this would imply is that we would need to add a try-except block to all the MACO extractors so way it doesn't break anything if they remain in the same project (this is doable and I don't see any problem with this).


> 4\. another thing that would be nice for you i guess, is splitted dependencies, to have extractor dependencies independent to main pyprojects so that would make smaller main pyproject, but main dependency i guess is yara y pycryptodomex for most of the cases. but you could have pure extractor related dependencies

This would be our most ideal scenario: where the extractors live in a separate project, specify their package dependencies, and could be submoduled/merged into CAPE upon installation. That way users that are just interested in running the extractors could just clone that one part rather than all of CAPE. 

But as you mentioned, right now they are pretty tight knit but I noticed that there are only 9 extractors that use `lib.cuckoo.common.*` so maybe that's something that could be pulled out and stored in this extractor-only repository? The other side of this is the YARA rule dependency that extractors have which could follow suit of being included in the extractor repo and then merged into the whole of CAPE.

---

An alternative measure I could see, which could be acceptable, is to move the `convert_to_MACO` within the CAPE extractors and change the output to return a dictionary rather than a Pydantic model. This removes the `maco` package dependency from the project entirely but this in some part moves the responsibility of the maintaining the conversion call in the main project (which for the most part, only people who call that function explicitly would be aware of like CCCS or anyone using Assemblyline.)

doomedraven:
yes is hard one, ok so lets do one thing. im gonna merge and later we can research how to properly do it properly. the main problem as you mentioned yaras and extractors. we love community contribution, but we also see that they not always stable/can be trusted due to bad data generation etc. So normally we keep CAPE stuff in core, and the rest outside. So split it into projects is a bit hard right now, but i guess we can move parsers into separate project under core and community and just load them as it is now will speak with kevin after his pto

cccs-kevin:
> i have a question, do you use this script? https://github.com/CAPESandbox/community/blob/master/utils/download_parsers.py

Yes we do use that script. @cccs-mog can confirm it's usage to download the latest parsers via a cronjob

doomedraven:
WIP https://github.com/CAPESandbox/CAPE-parsers

doomedraven:
Hello guys, we do have now it live on pypi, MACO wasn't fixed yet, but the parsers itself works just fine

https://pypi.org/project/CAPE-parsers/0.1.0/
image

so with having this as package, im considering more to move MACO to community and here is my reason, don't get me wrong i don't have nothing against MACO, but i neither write/maintain public parsers as we relay completely on our own parsers. So as im not maintaining them, if something changes, MACO formatting will be using by CCCS 99% so is easier to keep it fixed in community repo, than release new package right?

im open to proper discussion of this, i will fix all imports in MACO so they can import properly cape extractors, but if that is fine for you to put them in comunity, i will procede and we will be done very soon with this cape parsers split from capev2

PS what i don't see in this PR is invocation of the MACO by cape, or maybe there is too many files and it escapes from my eyes, so i guess you have code on your side right?

doomedraven:
what i would love to highlight in the cape-parsers `load_cape_parsers` is ability to load all(cape/community) or only one of those + support list of modules to exclude from loading, in case any parser produce bad data and you want to exclude it, specially if that is community module and nobody fixing it. i have fixed all MACO imports to use parsers from library, i still need to fix path to yaras, i think instead of having them in `modules/processing/parsers/MACO` i prefer to put them inside `modules/parsers/MACO` what do you think?

cccs-rs:
> so with having this as package, im considering more to move MACO to community and here is my reason, don't get me wrong i don't have nothing against MACO, but i neither write/maintain public parsers as we relay completely on our own parsers. So as im not maintaining them, if something changes, MACO formatting will be using by CCCS 99% so is easier to keep it fixed in community repo, than release new package right?

I think this will be fine if we can use the community repo and it contains the pip package as a dependency in the pyproject.toml and assuming everything we need is in that repository. I can do some testing when you think the repository is ready.

> PS what i don't see in this PR is invocation of the MACO by cape, or maybe there is too many files and it escapes from my eyes, so i guess you have code on your side right?

So I never really got around to that part since I know in the past the UI doesn't play well with rendering nested data which MACO is usually composed of. I believe previous discussions were that MACO could be an alternative view of the raw config but I don't know if that's still the case.



> i have fixed all MACO imports to use parsers from library, i still need to fix path to yaras, i think instead of having them in `modules/processing/parsers/MACO` i prefer to put them inside `modules/parsers/MACO` what do you think?

From our perspective if you move it closer to the root, it shouldn't make a difference. What's the strategy for resolving the yara dependencies atm if the extractors were moved to community? Would there be a sync process between the two?

cccs-rs:
I do want to say I really appreciate the efforts being made to help us out 😍

doomedraven:
yep it won't be merged till we find working solution for everyone here.
cape-parsers are part of capev2 pyproject, but i can put it there if you need. the idea is the same as now:
1. git pull CAPEv2
2. then fetch community repo, 
cape-parsers is inside of pyproject of CAPEv2, but i can put that in community as well if you need it, if you gonna use it with capev2, then you just use capev2's pyproject.

yes is just replace import path on your side i guess. i have fixed all imports in MACO, just need to fix yara path, yara stay in CAPEv2, as detection is done on CAPEv2, and cape invokes parsers after match, the extractor that needs their own yara, is embedded into extractor as it was before, so no really big change, just cleanup, flexibility etc

cccs-rs:
Would there be any opposition to embedding the YARA rules into the MACO extractors? This would resolve a path-relative dependency but still give the flexibility to use the YARA rules when running within or without CAPEv2.

what I'm thinking would be:
```python
try:
    yara_rule = open(os.path.join(os.path.dirname(__file__).split('/modules', 1)[0], f"data/yara/CAPE/{family}.yar")).read()
except FileNotFoundError:
    yara_rule = "SOME EMBEDDED RULE"
```

This would imply that at least for CCCS and MACO, we would need to update the rules ourselves when we do find there are problems and it wouldn't be a concern for CAPEv2 (unless these rule changes are in fact useful to CAPEv2 as well).

doomedraven:
yes you can do that, you can do anything to MACO files, as you will be probably uniq user of those. so do what best works for you. Yes the pitfall of embedding i guess is that you need to ensure that you update it if you use outside of cape or not reading the current yara which will be updated

check out, is the default maco, as imports only required update

https://github.com/CAPESandbox/community/pull/469

cccs-rs:
> CAPESandbox/community#469

I think this works for us, I'll do some testing later today before you merge that branch in, if that's okay 😁

doomedraven:
great, thank you for help, le me know about the dependencies, if you want them in community or it's used with capev2 pyprojects

PR #2379, opened 2024-10-30
Add Snake Keylogger config extractor

YungBinary:
Here's some previews, tested with lots of payloads :) if someone knows of a better way to get a field's value let me know, otherwise what I ended up doing was extracting all static strings and their associated field names, then using them later on when needing to lookup what the field value is for a specific field.

The PR for the Snake payload test file is: https://github.com/CAPESandbox/CAPE-TestFiles/pull/13

Screenshot from 2024-10-30 00-57-16

Screenshot from 2024-10-30 00-58-40

I'm sure in time the technique they're using to encrypt strings will change but for now this is how they do it, using the first 8 bytes of an MD5 calculated against a static string as the key with DES (ECB mode).

```
public static string decrypt_string(string string_30, string string_31)
{
	string result;
	try
	{
		DESCryptoServiceProvider descryptoServiceProvider = new DESCryptoServiceProvider();
		MD5CryptoServiceProvider md5CryptoServiceProvider = new MD5CryptoServiceProvider();
		byte[] array = new byte[8];
		byte[] sourceArray = md5CryptoServiceProvider.ComputeHash(Encoding.ASCII.GetBytes(string_31));
		Array.Copy(sourceArray, 0, array, 0, 8);
		descryptoServiceProvider.Key = array;
		descryptoServiceProvider.Mode = CipherMode.ECB;
		ICryptoTransform cryptoTransform = descryptoServiceProvider.CreateDecryptor();
		byte[] array2 = Convert.FromBase64String(string_30);
		string @string = Encoding.ASCII.GetString(cryptoTransform.TransformFinalBlock(array2, 0, array2.Length));
		result = @string;
	}
	return result;
}
```


kevoreilly:
Thanks for this! ❤️

PR #2380, opened 2024-10-30
maco optional

doomedraven:


doomedraven:
@cccs-rs you need to use ["maco"] as in king rat parsers. once you aware i will merge it

doomedraven:
btw https://github.com/CAPESandbox/CAPE-parsers, i still working on that and it will take some time to properly split them, but work started, lets see if i will have success on it

PR #2388, opened 2024-11-04
Update package for rdp package

cccs-mog:


doomedraven:
there is interesting tool from Will https://github.com/wmetcalf/rdp_holiday

PR #2398, opened 2024-11-14
feat: clean up cuckoo-sflock tmp folder

daniel-oronsi:
This PR adds `cuckoo-sflock` to the list of tmp folders for cleanup. I noticed it wasn’t included in when this function was changed a few months ago, and ran into issues on my CAPE instance, so thought it’d be helpful to add.

Thanks!


doomedraven:
thanks

PR #2415, opened 2024-11-28
Update cloning.rst

cccs-mog:


doomedraven:
thank you, i will add that configurable from the config tomorrow, so no need to change the disguise.py by hand, aftar that i will mode this and merge

doomedraven:
could you test the update plz before i merge?

doomedraven:
\c @dsecuma for headups

cccs-mog:
So I tested the upgrade and seem to work fine. Wondering what the goal of the config section gateways is ? Is it related ? Or maybe disguise should have his own section ?(Don't remember if the sections are for server side auxiliary) Small details but regardless of the presentation it work good now.

doomedraven:
ok, removed get as it useless as we have default value in config anyway, thank you going to merge

PR #2418, opened 2024-12-02
fix for interactive desktop analysis using rdp

boundprometheus:
This PR is a fix for using RDP for interactive desktop analyses. Without this change, I was unable to get RDP working.

There is also a new config for indicating if certificate verification of the remote desktop should be ignored. The cert verification only applies to connections over RDP; it is ignored for VNC.

### DIFF
--- a/conf/default/web.conf.default
+++ b/conf/default/web.conf.default
@@ -194,6 +194,7 @@ guest_width = 1280
 guest_height = 1024
 # rdp settings
 guest_rdp_port = 3389
+ignore_rdp_cert = false
 
 [packages]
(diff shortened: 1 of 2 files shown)


doomedraven:
thanks, i guess about certificate you mean https right?

boundprometheus:
Windows allows you to install a TLS cert for use with RDP. RDP clients will attempt to verify the signer back to the root certificate, but if validation fails (for example if the cert is self-signed), they will allow you to accept the warning and connect anyway. However, guacamole will fail the connection and report a TLS/SSL error, unless the `ignore_cert` argument is specified.

PR #2419, opened 2024-12-03
Update poetry binary location. Fix pip errors for ubuntu 22.04.

ChrisThibodeaux:
- Fixed all `poetry: command not found` errors.
- Kept the ubuntu 24.04 updates and fixed the pip errors with 22.04.

Likely a more graceful way of pointing to the Poetry binary than placing `/etc/poetry/bin/poetry` everywhere, but this fixes all of the install issues on ubuntu 22.04.

### DIFF
--- a/extra/libvirt_installer.sh
+++ b/extra/libvirt_installer.sh
@@ -2,7 +2,7 @@
 set -ex
 
 # run this via...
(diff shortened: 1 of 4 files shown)


doomedraven:
thanks

PR #2420, opened 2024-12-03
Improve trend unquarantine performance

josh-feather:
Prior to these changes `trend_unquarantine` was passing the entire sample through a XOR routine before checking the XOR'd file header to determine if it was a quarantined file.

For a 100MB sample it was taking ~22 seconds to XOR the file.

```
%Own   %Total  OwnTime  TotalTime  Function
0.00%   0.00%   21.58s    21.58s   bytearray_xor (lib/cuckoo/common/quarantine.py)

Note, `bytearray_xor` is called by `trend_unquarantine`
```

By changing the function to only XOR the first 10 bytes of the file (header) before checking the file signature, we remove significant amounts of processing time for what will likely be all samples being submitted.

Using the same sample, post-change, the whole `unquarantine` process now only takes 0.770s.

```
%Own   %Total  OwnTime  TotalTime  Function
0.00%   0.00%   0.040s    0.770s   unquarantine (lib/cuckoo/common/quarantine.py)

Note, `trend_unquarantine` is called by `unquarantine`
```

Ultimately, I'm not convinced that `trend_unquarantine` works anyway because:
 - there were syntax errors in the `.decode` functions.
 - `read_trend_tag` returns a `tuple[int, bytes]`, the tag code and the tag data. When unpacking the original filename tag (tag code 2), it passes the tag data (bytes) into the `str` func, encodes it and then decodes it again. This means the original filename is only ever a string representation of bytes (see below), causing the downstream functions to fail (in my testing, without a proper quarantined file).
 
```
>>> str(b"hello").encode("utf16").decode(errors="ignore")
"b\x00'\x00h\x00e\x00l\x00l\x00o\x00'\x00"  
>>> str("hello".encode("utf16")).encode("utf16").decode(errors="ignore")
"b\x00'\x00\\\x00x\x00f\x00f\x00\\\x00x\x00f\x00e\x00h\x00\\\x00x\x000\x000\x00e\x00\\\x00x\x000\x000\x00l\x00\\\x00x\x000\x000\x00l\x00\\\x00x\x000\x000\x00o\x00\\\x00x\x000\x000\x00'\x00"
```

### DIFF
--- a/lib/cuckoo/common/quarantine.py
+++ b/lib/cuckoo/common/quarantine.py
@@ -583,14 +583,14 @@ def trend_unquarantine(f):
     for _ in range(numtags):
         code, tagdata = read_trend_tag(data, offset)
         if code == 2:  # original filename
-            origname = str(tagdata).encode("utf16").decode(error="ignore").rstrip("\0")
+            origname = str(tagdata).encode("utf16").decode(errors="ignore").rstrip("\0")
         elif code == 6:  # base key
             basekey = struct.unpack("<I", tagdata)[0]
         elif code == 7:  # encryption method: 1 == xor FF, 2 = CRC method
             encmethod = struct.unpack("<I", tagdata)[0]
         """
         elif code == 1:  # original pathname
-            origpath = str(tagdata).encode("utf16").decode(error="ignore").rstrip("\0")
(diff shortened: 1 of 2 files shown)


doomedraven:
nice finding, thank you

PR #2427, opened 2024-12-18
check flare-capa with python version

dsecuma:
According to the CAPA release information (see https://github.com/mandiant/capa/releases/tag/v7.4.0), version 8.x of CAPA is currently compatible with the Python version used by CAPE.


doomedraven:
Thanks David, yes, we just have some small issue with GitHub ci, even saying run on py3.10 it installs it under 3.12, is why it not in dependencies right now

PR #2429, opened 2024-12-19
Enable IPv4 forwarding at rooter startup

seanthegeek:
Also add the `--sysctrl` option to the `rooter.py` CLI


doomedraven:
qq, why do you write 0? with just run command it should be more than enough right?

seanthegeek:
I'm guessing you meant 1 instead of zero. I thought the same thing, but the documentation says to do both, so I'm following the documentation.

https://capev2.readthedocs.io/en/latest/installation/host/routing.html

> $ echo 1 | sudo tee -a /proc/sys/net/ipv4/ip_forward
> $ sudo sysctl -w net.ipv4.ip_forward=1

It's also weird that the documentation is using `tee -a` to append to the file.  We probably don't need that at all. I'll do some tests.

seanthegeek:
Yep, only the `sysctl` command is needed. I removed the file edit from the PR.

doomedraven:
thanks for PR

PR #2446, opened 2025-01-03
Don't overwrite the 'target' key for every detonation.

tbeadle:
This fixes a regression introduced in https://github.com/kevoreilly/CAPEv2/commit/d896e8d50f70216ab9847b2880f63eff6828fab1 where every file detonation now has {"category": "url"} as the entire content of the "target" key. This removes all of the information about the target file.


kevoreilly:
Thank you

doomedraven:
damn, thanks Tommy, i did test but didn't get any problem, should have something wrong %)

PR #2449, opened 2025-01-07
Introducing wmi_etw auxiliary module

IridiumXOR:
During the analysis of a sample, we observed that it could modify the Registry contents, but Cape was unable to detect this activity (tracing WMI StdRegProv::CreateKey calls). After conducting a manual analysis, we discovered that the sample initiated a Registry modification request through WMI APIs, which Cape seems unable to properly track. To address this, we developed a plugin, *wmi_etw.py*, that enables the tracking of WMI activity via ETW. The code is based on the existing *dns_etw.py* plugin and saves the log trace in *aux/wmi_etw.json*. It is important to note that, like the *dns_etw.py* plugin, it requires the installation of the *pywintrace* Python package within the test VM to function correctly.

### DIFF
--- a/analyzer/windows/modules/auxiliary/wmi_etw.py
+++ b/analyzer/windows/modules/auxiliary/wmi_etw.py
@@ -1,0 +1,191 @@
+import json
+import logging
+import os
+import pprint
+from collections.abc import Iterable, Mapping
+
+from lib.common.abstracts import Auxiliary
+from lib.common.results import upload_to_host
+from lib.core.config import Config
+
+log = logging.getLogger(__name__)
+
+SAFELIST = []
+
+ETW = False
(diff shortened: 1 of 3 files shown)


kevoreilly:
Very interesting! Thank you. Any chance you could share the example sample mentioned? Or its hash!

IridiumXOR:
hash: 0a06a25efb1dfea94e24096f64b40cc3661ef97a0de194fd83d2ca8d9a3648cc
Our sample come from VT, but it seems available also from open-source repositories

doomedraven:
thank you very much, we need to add entry to aux config to load it, i will add it after the dinner

doomedraven:
\cc FYI @dsecuma

PR #2458, opened 2025-01-10
Update Arkei.yar Yara Rule

YungBinary:
Update Arkei.yar Yara Rule


kevoreilly:
Thanks @YungBinary I hope to get round to tightening the Stealc detection at some point! I will need to take a good look at how to differentiate between the two families.

PR #2459, opened 2025-01-13
restore behavior defaults to a dictionary

nbargnesi:
Prior to PR #2454, behavioral data defaulted to a dictionary pointing to a list. The PR changed it to an empty list. This is expected to be a dictionary everywhere, and most code expects the "processes" key to exist and either be empty or have process data in it.

The specific commit this changed in is https://github.com/kevoreilly/CAPEv2/commit/11fb7b1964ae0836e1a69ec3734c346594b08a2e.


kevoreilly:
Thanks Nick

PR #2468, opened 2025-01-27
Make the logging and exception handling inside of `processing_finished` more succinct

josh-feather:
Makes the logging and exception handling inside of `processing_finished` more succinct

- Merges the `pebble.ProcessExpired` and generic `Exception` clauses as they are functionally identical. I've retained code-level visibility of `pebble.ProcessExpired`, despite it not being technically required.
- Removes the recently added `exc_classname` and `exc_message` variables, introduced in commit 1556340. The exception name is not required in the case of the `TimeoutError` clause, as the exception message provides enough insight.  In the second, generic clause, `exc_info` ensures the exception type, traceback and stack are contained in the log message.
- Replaces the `Database()` initialization calls in the exception handlers with a reference to the already initialized database, `db`, instead.

### DIFF
--- a/utils/process.py
+++ b/utils/process.py
@@ -286,18 +286,11 @@ def processing_finished(future):
             _ = future.result()
             log.info("Reports generation completed for Task #%d", task_id)
         except TimeoutError as error:
-            exc_clsname = error.__class__.__name__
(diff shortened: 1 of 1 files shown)


doomedraven:
thank you i had it in my ToDo

PR #2474, opened 2025-01-30
fix litereport failure due to invalid utf-8 byte string

ClaudioWayne:
use chardet to avoid UnicodeDecodeError due to invalid utf-8 byte string.
Sample: 70524de53d60119c2913370232effaf1e551db9308909dba2cd1331bb0fa0f19


doomedraven:
interesting, thank you

PR #2482, opened 2025-02-03
Wrap signature init in a try/catch

josh-feather:
This prevents the processing step from failing if there is a problem in a signatures `__init__` method.


doomedraven:
Is there any issue or is just to prevent possible fails?

josh-feather:
> Is there any issue or is just to prevent possible fails?

This will prevent regressions like [this](https://github.com/CAPESandbox/community/commit/9ef92fb942b181a212cf4d4c95bf41d1e795b4fd) from killing the processor

PR #2498, opened 2025-02-20
A couple of ruff %s fixes were missed

rkoumis:
- The change was from `"%s" % str(id)` to `"%s", str(id)`
- Should have been just `str(id)`
- Follows on from f89c890143b59bdbcc9b9c8b5e15fbc5fd827afd
- also changed a comment re: integrations.conf renaming


kevoreilly:
Thank you

PR #2502, opened 2025-03-01
Replace apt-get with apt and enforce sudo and -y usage in installer/cape2.sh

t-mtsmt:
Updated installer/cape2.sh to replace apt-get with apt for consistency and ensured all package commands use sudo and -y.


doomedraven:
hello, thank you, that is how i had script initially, but `apt` then throw warning all the time so i had to change it back to `apt-get`. but the part of `-y` is good

t-mtsmt:
Thank you for your feedback and for updating it back to apt-get. I appreciate it!

PR #2511, opened 2025-03-05
Fix downloaders order

enzok:


doomedraven:
thank you, can we test that tests are useless now? :D

PR #2531, opened 2025-03-21
Fix and update kvm-qemu.sh

ClaudioWayne:


ClaudioWayne:
Ups should be:
LIB_VERSION=11.1.0
libvirt_version=11.1.0

PR #2554, opened 2025-04-09
useragent switcher for firefox

dsecuma:
Change the Firefox user agent for URL analysis.

The user agent must be sent as a base64-encoded string.
Provide the path to your current Firefox user profile, where the pref.js file is located.


doomedraven:
thank you

PR #2556, opened 2025-04-10
cleanup

doomedraven:


dev_utils/mongodb.py
```diff
@@ -186,38 +186,53 @@ def mongo_drop_database(database: str):
...
+            mongo_delete_many("analysis", {"info.id": {"$gt": range_start}})
+        elif range_end:
+            mongo_delete_many("analysis", {"info.id": {"$lt": range_end}})
+        else:
+
+            mongo_delete_many("analysis", {"info.id": {"$in": task_ids}})
+        mongo_delete_calls(task_ids=task_ids, range_start=range_start, range_end=range_end)
         if task_ids:
             for hook in hooks[mongo_delete_data]["analysis"]:
                 hook(task_ids)
```
josh-feather, line 213: Hmm... I think the changes made in https://github.com/kevoreilly/CAPEv2/pull/2556/files need to be reverted. I'm not sure the more complex function signatures (`mongo_delete_data` and `mongo_delete_calls`) add value and the Mongo hooks only take the `task_ids` parameter, which hasn't been accounted for.

josh-feather: Consider doing something like this instead:
https://github.com/kevoreilly/CAPEv2/pull/2567/files

Note, these changes are untested.

PR #2566, opened 2025-04-19
fix: update cleanup logic to use direct config values

MHM98:
Hello,
This PR try to solve some issues that i think coming from in #2556 and caused cleaner logic to does not work properly.

### DIFF
--- a/lib/cuckoo/common/cleaners_utils.py
+++ b/lib/cuckoo/common/cleaners_utils.py
@@ -117,11 +117,11 @@ def free_space_monitor(path=False, return_value=False, processing=False, analysi
         "delete_mongo": config.cleaner.mongo,
     }
     if config.cleaner.binaries:
-        cleanup_dict["delete_binaries_items_older_than"] = convert_into_time(config.cleaner.binaries)
+        cleanup_dict["delete_binaries_items_older_than"] = config.cleaner.binaries
     if config.cleaner.tmp:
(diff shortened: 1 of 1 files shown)


doomedraven:
that is wrong, config should have 12h etc format, so it won't bbe able to convert to proper format without that

MHM98:
this is my cleaner configuration in cuckoo.conf file

> [cleaner]
> enabled = yes
> binaries = 1h
> tmp = 1h
> analysis = 1h
> mongo = yes
> unused_files_in_mongodb = yes

when i hit my limitation configured  by `freespace = 20000`   the **cape.service**  service crash and logs show `Invalid time range format` error.
I troubleshoot the issue inside `CAPEv2/lib/cuckoo/common/cleaners_utils.py` file and realize `convert_into_time` function called twice with wrong value. (for example once [here](https://github.com/kevoreilly/CAPEv2/blob/563fd244405d62ca08793c67bf5ca80431a58185/lib/cuckoo/common/cleaners_utils.py#L124)  in `free_space_monitor` function and again inside `cuckoo_clean_before` function   in [here](https://github.com/kevoreilly/CAPEv2/blob/563fd244405d62ca08793c67bf5ca80431a58185/lib/cuckoo/common/cleaners_utils.py#L491) )
I made the changes in this PR and it's look it solve the problem

doomedraven:
ah ok yes make sense, i did check now properly and see what you saying, thank you

PR #2567, opened 2025-04-21
Fix regressions in mongo_delete_data, improve mongo_delete_calls

josh-feather:
- Use `None` as a default parameter for `task_ids`.
- Instead of executing delete queries using the provided range, use the provided range to pull task_ids from Mongo to allow the hooks to execute properly.


josh-feather:
Note, this is completely untested.

PR #2577, opened 2025-04-29
Wireshark installation fails in non-interactive environments

sunba91-su:
### Problem
When running the installer script, the terminal exits unexpectedly during the Wireshark package installation. This is because wireshark-common prompts for user input (e.g., permission to allow non-superusers to capture packets), which requires an interactive terminal.

This breaks automated installations or any environment (like CI/CD pipelines or Docker builds) where user interaction is not possible.

### Solution
Set the environment variable DEBIAN_FRONTEND=noninteractive before installing Wireshark to suppress the interactive prompt:

```bash
DEBIAN_FRONTEND=noninteractive apt-get install -y wireshark
```

This ensures the script runs smoothly in non-interactive sessions.

### Additional Notes

- The script could optionally pre-answer debconf selections if needed.

- Tested on Debian/Ubuntu systems.

- Does not impact systems where interaction is possible — only improves automation compatibility.


doomedraven:
thank you, maybe this should be set on top of the install script to apply to all apt invocations?

sunba91-su:
You're absolutely right — setting `DEBIAN_FRONTEND=noninteractive` globally at the top of the install script is a better approach. This ensures all `apt-get` commands behave consistently in non-interactive environments, not just the Wireshark installation.

I'll update the script accordingly to export this environment variable near the top. Thanks for the feedback!

PR #2596, opened 2025-05-11
FIX: replace <WOOT> with proper env in replace_seabios_clues_public

MHM98:
There was some place in `kvm-qemu.sh` script that  the `<WOOT>` was not replaced with proper value.
So i defined an env and replace them accordingly


doomedraven:
hello, this is not fully correct, you replace 2 different patterns with just 1, they should have their own replacement patterns

doomedraven:
thank you

PR #2603, opened 2025-05-20
Update database.py so that the sample database will use bigint for fi…

markykchua:
…le_size.

For users that have an existing database when upgrading CAPEv2, you need to alter the database schema before the upgrade: 

ALTER TABLE samples
ALTER COLUMN file_size TYPE BIGINT;

And restart the related services.

### DIFF
--- a/lib/cuckoo/core/database.py
+++ b/lib/cuckoo/core/database.py
@@ -44,6 +44,7 @@ try:
         ForeignKey,
         Index,
         Integer,
+        BigInteger,
         String,
         Table,
         Text,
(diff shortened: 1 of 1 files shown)


doomedraven:
hello thank you, the fix is partial, we need alembic migration to properly make users aware of it, we can't just hope that they will check readme. i will implement it once im back in june

markykchua:
> hello thank you, the fix is partial, we need alembic migration to properly make users aware of it, we can't just hope that they will check readme. i will implement it once im back in june

perfect, thanks a lot!

doomedraven:
this is now addressed in upcoming database modernization/cleanup in staging branch, thanks for headups

PR #2636, opened 2025-07-04
sqlalchemy 2

doomedraven:


josh-feather:
Will the existing migrations have to be updated to support SQLAlchemy 2.0?

doomedraven:
good question, i have dropped all less 3 last migration, but forgot that it might require sqlalchemy 2 update, i will have to see how hard it will be to port to it, if it to hard, no, as i still fixing issue on parent<>child relation, as implementation is wrong/bad

josh-feather:
> good question, i have dropped all less 3 last migration, but forgot that it might require sqlalchemy 2 update, i will have to see how hard it will be to port to it, if it to hard, no, as i still fixing issue on parent<>child relation, as implementation is wrong/bad

Assuming you've seen the [upgrade guide](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html)?

doomedraven:
i saw it, but did read in diagonal only

PR #2653, opened 2025-07-27
Keep file when demux adds default platform

itamarga:
If demux_sample returns (path, "windows") for a plain file (e.g. a csv named SAMPLE.XLS), tuple-based check saw:
(path, "") != (path, "windows")
and deleted the upload, triggering “Extracted file doesn’t exist”. Compare paths only.

### DIFF
--- a/conf/default/reporting.conf.default
+++ b/conf/default/reporting.conf.default
@@ -147,10 +147,6 @@ user = admin
 pass = admin
 realm = Moloch
 
-# Community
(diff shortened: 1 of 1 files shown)


doomedraven:
hello, thank you, can you apply Gemini suggested improvement?

itamarga:
Done :+1:

doomedraven:
thank you, sorry for delay on review

PR #2656, opened 2025-07-30
Add PolarProxy plugin for TLS MiTM

aaron-boyd:
This merge request adds the capability to run PolarProxy on a per task basis to man-in-the-middle TLS connections from target VMs on a predefined port.

## Install 

```
$ ./cape2.sh polarproxy cape
```

## Usage

Use `polarproxy=1` option during submission. The MiTMed port can be changed using the `tlsport=<PORT>` option.

## Functionality

The rooter shims a `REDIRECT` rule at the top of the `PREROUTING` table to intercept any traffic from a VM destined for port 443 and redirects it to PolarProxy listening on an ephemeral port. The traffic is decrypted and written to `storage/analyses/XX/polarproxy/tls.pcap`. During the processing step, this plugin will use `mergecap` to stitch the plaintext TLS traffic back into the PCAP generated from `tcpdump`. This modified PCAP is available just like normal through the UI/API. The TLS only PCAP is available through the `/apiv2/tasks/get/tlspcap/XX/` endpoint.

PolarProxy includes a TLS firewall. Using the `bypass_list` and `block_list` option in `polarproxy.conf`, a user can define a list of domain regular expressions to block connections to (ie Windows telemetry) or allow the connection to go through but not MITM it.


doomedraven:
thank you very much!

PR #2665, opened 2025-08-11
Reduce vivisect.base and vivisect.impemu cape-processor logs

ChrisThibodeaux:
Cape-processor logs are flooded with vivisect logs. These changes keep relevant `vivisect` logs and remove `vivisect.base` and `vivisect.impemu` logs below the `ERROR` level.

### DIFF
--- a/lib/cuckoo/common/integrations/capa.py
+++ b/lib/cuckoo/common/integrations/capa.py
@@ -50,9 +50,12 @@ if integrations_conf.flare_capa.enabled:
             from pydantic_core._pydantic_core import ValidationError
 
             # Reduce vivisect logging verbosity
-            logging.getLogger("vivisect").setLevel(logging.CRITICAL)
+            logging.getLogger("vivisect").setLevel(logging.INFO)
             logging.getLogger("vivisect.base").setLevel(logging.CRITICAL)
             logging.getLogger("vivisect.impemu").setLevel(logging.CRITICAL)
+            logging.getLogger("vivisect.parsers").setLevel(logging.CRITICAL)
+            logging.getLogger("vivisect.tools").setLevel(logging.CRITICAL)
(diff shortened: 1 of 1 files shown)


doomedraven:
interesting, i had disable it completely

ChrisThibodeaux:
I found that the current settings had no impact on logs at `INFO` level or above. I'm not entirely sure how `NOTSET` works, but it seems to use a lower-heirarchy logger's level to set the one in capa.py

doomedraven:
Captura de pantalla 2025-08-11 a las 23 22 05

maybe better to set it to critical?

ChrisThibodeaux:
I can definitely make that change, not a problem.

I found some of the `[vivisect]` logs to be useful, such as:
```
2025-08-11 15:39:37,257 [Task 894] [vivisect] INFO: Percentage of discovered executable surface area: 95.8% (427791 / 446464)
2025-08-11 15:39:37,257 [Task 894] [vivisect] INFO:    Xrefs/Blocks/Funcs:                             (31491 / 23131 / 1610)
2025-08-11 15:39:37,257 [Task 894] [vivisect] INFO:    Locs,  Ops/Strings/Unicode/Nums/Ptrs/Vtables:   (89714:  83169 / 623 / 782 / 1339 / 620 / 0)
```

However, leaving it at `INFO` _does_ still output more logs than are useful. Particularly, `vivisect.parsers.pe` is still a pain and really overkill with logging..

I agree on the swap to `CRITICAL` and have tested it locally.

doomedraven:
we can mute per subparsers if we want that is not a big problem. so maybe we should add `logging.getLogger("vivisect.parsers.pe").setLevel(logging.CRITICAL)` ? could you test it please

ChrisThibodeaux:
No problem, I will test that out today

doomedraven:
Thank you

PR #2667, opened 2025-08-12
Update usages with expected location of 7zz binary

ChrisThibodeaux:
Replace the leftover code spots pointing to `/usr/bin/7z`.

@doomedraven 
My preferred fix:
1. Set `7zz` binary relative location in only one place in one config file.
2. Create a constant using `SEVENZIP_BIN = os.path.join(CUCKOO_ROOT, <config_value>)` that can be imported where needed.

Let me know if I should do that instead and I will make the changes to this PR.

### DIFF
--- a/lib/cuckoo/common/integrations/file_extra_info.py
+++ b/lib/cuckoo/common/integrations/file_extra_info.py
@@ -119,6 +119,9 @@ if integration_conf.SevenZip_unpack.binary:
     tmp_sevenzip_binary = os.path.join(CUCKOO_ROOT, integration_conf.SevenZip_unpack.binary)
     if path_exists(tmp_sevenzip_binary):
         sevenzip_binary = tmp_sevenzip_binary
+# fallback
+if not path_exists(sevenzip_binary):
+    sevenzip_binary = "/usr/bin/7z"
 
 if processing_conf.trid.enabled:
     trid_binary = os.path.join(CUCKOO_ROOT, processing_conf.trid.identifier)


modules/processing/suricata.py
```diff
@@ -396,7 +396,7 @@ def run(self):
                     log.warning("Unable to delete suricata file subdirectories: %s", e)
 
         if SURICATA_FILES_DIR_FULL_PATH and path_exists(SURICATA_FILES_DIR_FULL_PATH) and Z7_PATH and path_exists(Z7_PATH):
-            # /usr/bin/7z a -pinfected -y files.zip files-json.log files
+            # /opt/CAPEv2/data/7zz a -pinfected -y files.zip files-json.log files
             cmdstr = f"cd {self.logs_path} && {Z7_PATH} a -p{FILES_ZIP_PASS} -y files.zip {SURICATA_FILE_LOG} {SURICATA_FILES_DIR}"
```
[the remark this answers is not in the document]
ChrisThibodeaux: Fixed

lib/cuckoo/common/integrations/file_extra_info.py
```diff
@@ -114,7 +114,7 @@
     unautoit_binary = os.path.join(CUCKOO_ROOT, integration_conf.UnAutoIt_extract.binary)
 if integration_conf.Inno_extract.binary:
     innoextact_binary = os.path.join(CUCKOO_ROOT, integration_conf.Inno_extract.binary)
-sevenzip_binary = "/usr/bin/7z"
+sevenzip_binary = os.path.join(CUCKOO_ROOT, "data/7zz")
```
doomedraven, line 117: hey, this was set in this way to prevent issues, as i set it as you setting it and it generated issues to some users, i will update it to fallback to old path if 7zz doesnt exist. apart of that. the rest looks fine

doomedraven:
thank you

lib/cuckoo/common/integrations/file_extra_info.py
ChrisThibodeaux: Makes sense. I should have made that backwards compatible. Thank you

PR #2674, opened 2025-08-19
Update Stealc.yar

bartblaze:
This should get rid of the FPs but not produce FNs.

### DIFF
--- a/data/yara/CAPE/Stealc.yar
+++ b/data/yara/CAPE/Stealc.yar
@@ -1,3 +1,4 @@
+import "pe"
 rule Stealc
 {
     meta:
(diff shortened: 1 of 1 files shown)


kevoreilly:
I prefer to improve the signature patterns rather than add exclusions.

Please supply details and hashes for the FPs.

bartblaze:
Agreed with this and had the same thought originally - although, that might mean getting rid of the $snippet1 string.. Example FP hashes:

03f891ce963ca916c865b91da8e9cd47dc97406c2cbc7d019964fe5b61d7343f
444ed825f744a8929d05f7900a8768e968e25489dfa0f73326833bc3bb5a55d9

image

Hope this helps,

PR #2683, opened 2025-08-25
Improve logging

Aeris1One:
Print full traceback instead of only the last occuring error (which is often caused by previous ones)
As an example, errors thrown by machineries are stripped of their details before being printed to the user (as seen on #2682).


doomedraven:
thank you

PR #2694, opened 2025-09-08
Prevent mouse emulation on Windows VMs from clicking on Console windows

josh-feather:
This avoids situations where the console is clicked, causing select-mode to be enabled and the process to be suspended. This ultimately leads to the timeout being hit and a detonation with little useful telemetry.

### DIFF
--- a/analyzer/windows/modules/auxiliary/human.py
+++ b/analyzer/windows/modules/auxiliary/human.py
@@ -147,6 +147,30 @@ def get_cursor_position():
     return {"x": pt.x, "y": pt.y}
 
 
+def cursor_over_console_window():
+    # Get current cursor position
+    pt = wintypes.POINT()
+    USER32.GetCursorPos(byref(pt))
+
+    # Get the window handle at the cursor position
+    hwnd = USER32.WindowFromPoint(pt)
+
+    if not hwnd:
+        return False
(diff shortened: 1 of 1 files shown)


kevoreilly:
Hi Josh, thanks for this, I'm testing this currently. Do you happen to have any good examples to test with?

josh-feather:
> Hi Josh, thanks for this, I'm testing this currently. Do you happen to have any good examples to test with?

I tested it with a full screen console app (with/without the fix applied) - I suspect that's probably the easiest way.

dsecuma:
Hi, It would be interesting to have the ability to enable or disable the populate_clipboard() function to prevent the clipboard from being overwritten with real content. For instance, clickfix malware.

kevoreilly:
Not a bad idea, but to play devil's advocate would your example be 'fixed' simply by disabling automated interaction?

josh-feather:
> Hi, It would be interesting to have the ability to enable or disable the populate_clipboard() function to prevent the clipboard from being overwritten with real content. For instance, clickfix malware.

The `populate_clipboard` function is only called once to prevent the copy buffer from being empty during analysis. Disabling human interaction completely is probably the easiest way to stop that from happening.

PR #2704, opened 2025-09-21
Use universal SID instead of locale-specific group names

AndreasHanikel:
Fixes agent failures on non-English Windows by using the SID  "S-1-5-32-545" instead of hardcoded "BUILTIN\Users" in icacls commands.

https://learn.microsoft.com/en-us/windows/win32/secauthz/well-known-sids


kevoreilly:
Danke!

PR #2710, opened 2025-09-25
Only show interactive options and infos when enabled and used (AI imp…

ClaudioWayne:
…rovements)
- Show interactive and manual check boxes depending on whether they are configured on submit page
- Display remote session information on the task status website only if it has been configured and activated.
- Disable Manual checkbox on submit page when interactive checkbox is unchecked
- AI improvements included


doomedraven:
AI doesn't complain about AI code that's great. Thank you

PR #2725, opened 2025-10-21
start libvirt VM in case snapshot was taken in shut-off mode

henry-spanka:
If the Libvirt VM Snapshot was not taken while the VM was running (shutoff) then the VM will not be automatically started which causes the job to timeout.

This PR checks the snapshot state and will start the VM if needed.

Newer libvirt specs (e.g. Windows 11 machine with pflash) do not support live snapshots, therefore this is particularly useful for these configuration where snapshots in shutoff state are required.

### DIFF
--- a/lib/cuckoo/common/abstracts.py
+++ b/lib/cuckoo/common/abstracts.py
@@ -453,7 +453,7 @@ class LibVirtMachinery(Machinery):
                 self.vms[label].revertToSnapshot(snapshot, flags=0)
                 _start_vm_if_necessary(snapshot, vm)
             except libvirt.libvirtError as e:
-                msg = f"Unable to restore snapshot {vm_info.snapshot} on virtual machine {label}. Your snapshot MUST BE in running state!"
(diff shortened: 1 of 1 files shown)


doomedraven:
Snapshot must be in running state! Is not optional

henry-spanka:
Can you give a short explanation why this is necessary?

doomedraven:
Because is how this project works, otherwise you waste a lot of time and resources on booting your vm

henry-spanka:
I understand that live snapshots are preferred but I don't see a downside to also allow to use offline snapshots as it's not always possible to create live snapshots.

Boot time in my case is about 20 seconds for Windows until agent registers which is fine for an installation with few users.

But if it's out of scope for this project then keep the PR closed. Just thought it might be useful but I'm a newbie in this project.

doomedraven:
you could search in issues and i think PRs, we spoke few times about that already. It would be ok if that wouldn't confuse users. So as docs says, it must be in running state. so follow docs ;)

PR #2727, opened 2025-10-23
Generate .pcapng files on-demand

josh-feather:
Generating .pcapng files during processing results in double the amount of network data being stored on disk. This is because the original .pcap is still required for reprocessing. This change introduces functionality that enables .pcapng files to be generated when the file download is attempted (when the file is required).

This brings two benefits:
1. PCAP-NG files aren't generated for every detonation, reducing the amount of disk space used.
2. Reduces processing time.

It's worth noting that the tradeoff is that it can be quite expensive to generate the .pcapng in the web thread if the PCAP is significant in size.


doomedraven:
hey nice PR thank you, but instead of utils y personally would suggest to put it under `lib/cuckoo/common/pcap_utils.py`

josh-feather:
> hey nice PR thank you, but instead of utils y personally would suggest to put it under `lib/cuckoo/common/pcap_utils.py`

Sure - I will do that.

doomedraven:
nice, thank you

kevoreilly:
Much appreciated 👍

PR #2753, opened 2025-11-19
Fix a bug that prevents terminal status from being reported by the agent

josh-feather:
Updates the POST /status endpoint to unset the async subprocess if the new status is terminal. This makes GET /status report the final analysis state, rather than the child process state.

See https://github.com/kevoreilly/CAPEv2/compare/master...josh-feather:CAPEv2:agent-terminal-status-update?expand=1#diff-14871e8d8c7b1cfa74aec39f46895f82101d3e070dc811eed3a6d1dd3e70e65dL446 for context

### DIFF
--- a/agent/agent.py
+++ b/agent/agent.py
@@ -92,6 +92,8 @@ class Status(enum.IntEnum):
         return None
 
 
+TERMINAL_STATUSES = [Status.COMPLETE, Status.FAILED, Status.EXCEPTION]
+
 AGENT_BROWSER_EXT_PATH = ""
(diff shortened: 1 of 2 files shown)


doomedraven:
@dsecuma any thougts here? it affects linux analyzer

dsecuma:
> @dsecuma any thougts here? it affects linux analyzer

I've just tried it on our Ubuntu configuration and it works fine. Perhaps some detonations could cause a forced timeout, but LGTM.

doomedraven:
Thanks for feedback

PR #2771, opened 2025-12-06
Fix dump memory

nalves599:


doomedraven:
@dsecuma get a look on this please when you will can

PR #2786, opened 2025-12-12
Replace cgi.FieldStorage with email/urllib for form parsing

doomedraven:
Refactors MiniHTTPRequestHandler to use the email and urllib modules for parsing multipart/form-data and application/x-www-form-urlencoded requests, replacing the deprecated cgi.FieldStorage. This improves compatibility with newer Python versions and handles file uploads and form fields more robustly. Also updates server shutdown logic to use the shutdown() method.

### DIFF
--- a/agent/agent.py
+++ b/agent/agent.py
@@ -112,11 +112,28 @@ state = {
 class MiniHTTPRequestHandler(http.server.SimpleHTTPRequestHandler):
     server_version = "CAPE Agent"
 
-    def _parse_form(self):
+    def _init_request(self, method):
         request.client_ip, request.client_port = self.client_address
         request.form = {}
         request.files = {}
+        request.method = method
 
+    def _decode_bytes(self, b):
+        try:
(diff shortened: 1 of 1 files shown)


chrisshaiman:
Hit this bug on a fresh CAPEv2 deployment with Python 3.12.10 (x86, Windows 11 guest). The store, execute, and execpy endpoints all fail with TypeError: write() argument must be str, not bytes from cgi.FieldStorage.make_file() creating text-mode tempfiles for binary data.

Tested the email.parser.BytesParser approach from this PR on Python 3.12.3 (Ubuntu 24.04 host) and 3.12.10 x86 (Windows 11 guest). Multipart form parsing works correctly — text fields decode to strings, file payloads preserved as bytes, no cgi import needed. This also unblocks upgrading the guest agent to Python 3.13+ since the cgi module is fully removed there.

This is blocking sample submission on our sandbox. Would love to see this merged. Thanks for the fix!

PR #2791, opened 2025-12-15
Fix the issue of network ports being empty in the report

B3i4hu:
The original location network port has not been obtained, which will cause the port information to remain empty.

After modification:
image


doomedraven:
interesting finding, thank you

PR #2833, opened 2026-01-08
css updates to bootstrap5

enzok:


doomedraven:
Can you also pr the test fix plz?

PR #2966, opened 2026-04-08
Add xdg-open package for generic graphical sample detonation on Linux

RobertBendun:
[`xdg-open`](https://man.archlinux.org/man/xdg-open.1.en) is a generic way to open files on linux that respects users preferences, making detonation of documents like pdf files in system configured pdf reader possible


kevoreilly:
Thank you

PR #2979, opened 2026-04-16
qickfix for  "false" route handling

MMikePL:
If a user or an API call set the routing option to "false":
1.Initial identification (line 540) worked: The system use four different values as indicators to disable routing: "none", "None", "drop", and "false" 
2.The "Drop" command was skipped: The code failed to call rooter("drop_enable", ...) because "false" was missing from the second check. This meant the network traffic might not have been explicitly blocked by the rooter as intended.
3.Cleanup was skipped: During task teardown, rooter("drop_disable", ...) was not called, potentially leaving the network state inconsistent for future tasks.

this fix ensure that rooter correctly enables and disables traffic dropping when "false" is used.

### DIFF
--- a/lib/cuckoo/core/analysis_manager.py
+++ b/lib/cuckoo/core/analysis_manager.py
@@ -606,7 +606,7 @@ class AnalysisManager(threading.Thread):
             )
             self.rooter_response = rooter("libvirt_fwo_enable", self.machine.interface, self.machine.ip)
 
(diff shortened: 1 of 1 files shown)


doomedraven:
https://github.com/kevoreilly/CAPEv2/commit/3673e80c1c0ea8c0576c8000c7e191d97f27d53d

PR #2980, opened 2026-04-16
Add more paths for Chrome & Chromium packages

adepasquale:


adepasquale:
None of the gemini suggestions make sense:

https://github.com/kevoreilly/CAPEv2/blob/f7a5d7f620bb98f913f0c270e5ee82c27396e9f4/analyzer/windows/lib/common/abstracts.py#L112-L114

PR #2987, opened 2026-04-20
Process-to-network attribution for sandbox analyses

wmetcalf:
## Summary

Tie every network artifact CAPE captures (suricata alerts/tls/http/files, network.tcp/udp/dns/hosts) back to the originating Windows process so analysts and downstream signatures can answer "which process did this?" without manual correlation.

A single `AttributionIndex` consumes six attribution sources (Sysmon EID 1/3/22, Microsoft-Windows-Kernel-Network ETW, DNS-Client ETW, sigma matched events) and exposes one query API used by every per-target enrichment loop. 5-tuple matching with `src_port` disambiguates multi-process flows to the same destination.

### What's added

- **Analyzer aux** `network_etw.py` — captures TCP/UDP connect events from `Microsoft-Windows-Kernel-Network` with periodic full-snapshot uploads. Off by default (`auxiliary.conf.default` — opt-in with `[Network_ETW] enabled = yes`).
- **Stop-priority** `-20` on `Network_ETW` and `Evtx` aux modules so they shut down AFTER capemon-related auxiliaries — late-fire C2 callbacks that fire between the analysis-stop signal and VM teardown still get captured + attributed.
- **Result-server transport** — `RESULT_UPLOADABLE` allowlist tightens path-traversal protection; new `is_replaceable_result_upload()` helper allows periodic re-uploads (tlsdump.log, dns_etw.json, network_etw.json, wmi_etw.json, sslkeylogfile/sslkeys.log) to truncate-write instead of silently failing with EEXIST after the first upload.
- **Decryption pipeline** — `decryptpcap.py` wraps gogorobocap to produce `dump_decrypted.pcap` + `dump_mixed.pcap`. `network.py` and `suricata.py` honour a new `pcapsrc` config knob (`auto`/`original`/`mixed`/`decrypted`) so users explicitly choose which pcap variant to analyse.
- **UI** — `_suricata_http.html`, `_suricata_files.html`, `_hosts.html` render process attribution columns/rows, gated on the existing `NETWORK_PROC_MAP` setting. `_hosts.html` shows multi-process attribution as badges with hover tooltips that explain the attribution chain. Pre-existing "asn cell skipped when empty causes column shift" bug fixed in the same edit.
- **Tests** — `tests/test_network_capture_integration.py` covers `AttributionIndex` build + query, `pcapsrc` resolution, and replaceable-upload behaviour.

### Source priority (highest confidence first)

1. Sysmon EID 3 (NetworkConnect) — full image path + 5-tuple
2. Microsoft-Windows-Kernel-Network ETW — live-captured periodic uploads
3. Sigma EID 3 matched events — tertiary catch
4. DNS-Client ETW × resolution data (suricata.dns / network.dns / network.hosts / sigma EID 22 QueryResults) — fallback for IPs we never saw a direct connect for. Avoids the "everything routes to svchost (dnscache)" failure mode.
5. Sysmon EID 22 (DnsQuery) — covers queries that fired before the DNS-ETW auxiliary subscribed
6. Sysmon EID 1 (ProcessCreate) — names processes that aren't monitored by capemon but show up in connection/DNS data

## Test plan

- [x] `pytest tests/test_network_capture_integration.py` — 7 passed, 0 failed
- [x] Live reprocess of a representative analysis — attribution counts: alerts 2/2, tls 1/1, http 10/10, files 12/13, tcp 3/3, dns 1/1, hosts 1/1
- [x] No regression on tasks without sysmon/Kernel-ETW data — falls through to DNS-based attribution where available
- [x] Path-traversal hardening verified via `RESULT_DIRECTORIES` allowlist
- [x] All three `cape`/`cape-web`/`cape-processor` services restart cleanly with the new code

### DIFF
--- a/analyzer/windows/modules/auxiliary/evtx.py
+++ b/analyzer/windows/modules/auxiliary/evtx.py
@@ -12,6 +12,15 @@ log = logging.getLogger(__name__)
 
 
 class Evtx(Thread, Auxiliary):
+    # Stop AFTER capemon-related auxiliaries so the final EVTX snapshot
+    # captures sysmon events from late-fire callbacks that fire between
+    # the analysis-stopping signal and the VM teardown (e.g. C2 callbacks
+    # the malware schedules after a delay). Without this priority bump,
+    # those events happen after the last EVTX snapshot and never reach
+    # the host-side processing modules.
+    start_priority = 0
+    stop_priority = -20
+
     evtx_dump = "evtx.zip"
 
     # Event log channels to collect
--- a/analyzer/windows/modules/auxiliary/network_etw.py
+++ b/analyzer/windows/modules/auxiliary/network_etw.py
@@ -2,7 +2,6 @@ import json
 import logging
 import os
 import shutil
-import socket
 import time
 from threading import Thread
 
@@ -16,7 +15,6 @@ from lib.common.etw_utils import (
     ProviderInfo,
     GUID,
     et,
-    encode,
 )
 
 log = logging.getLogger(__name__)
--- a/analyzer/windows/modules/auxiliary/network_etw.py
+++ b/analyzer/windows/modules/auxiliary/network_etw.py
@@ -95,7 +95,7 @@ if HAVE_ETW:
                 return True
             if dst_port in self._filter_ports or src_port in self._filter_ports:
                 return True
-            if dst_ip in ("127.0.0.1", "::1", "0.0.0.0", ""):
+            if dst_ip in ("127.0.0.1", "::1", "0.0.0.0", "::", ""):
                 return True
             return False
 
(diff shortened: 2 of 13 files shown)


wmetcalf:
Heads up on a force-push: I noticed after the initial review that my fork (`wmetcalf/CAPEv2`) was 93 commits behind `master`, which meant this PR's diff was misleading — in particular it showed `modules/processing/decryptpcap.py` as a new file when upstream already has one, and the diffs on `evtx.py` / `resultserver.py` were against a stale base.

I've now:
1. Fast-forwarded my fork's master to current upstream `master` (31be9978).
2. Rebased this branch on the synced master.
3. Resolved the `decryptpcap.py` add/add conflict by keeping upstream's existing file intact and only adding the three top-level helpers (`_get_option`, `_is_usable_pcap`, `resolve_processing_pcap_path`) that `network.py` / `suricata.py` import.
4. Removed a test that relied on a stricter `resultserver.py` allowlist refactor that isn't in this PR (the stricter refactor is defensible as a separate hygiene PR if you want it; not bundling it here).

The functional scope of the PR is unchanged. The diff is now a clean minimum against upstream. Sorry for the noise — will diff against upstream before opening future PRs.

PR #2994, opened 2026-04-24
apiv2: add download endpoints for pcap variants, TLS keys, ETW logs, bulk archives

wmetcalf:
## Summary

The web UI file-download view has long exposed `dump_decrypted.pcap`, `dump_mixed.pcap`, pcap bundles, pcapng (with embedded keys), tlsdump keylog, and whole-directory archives, but the REST API only served `dump.pcap` plus a `tasks_tlspcap` endpoint hard-coded to `polarproxy/tls.pcap` — a path no longer produced since CAPE moved to SSLproxy + GoGoRoboCap. Programmatic consumers currently have no way to pull the decrypted / merged pcaps or any of the ETW-based telemetry that now drives process attribution.

This PR brings apiv2 to parity with the web UI and adds endpoints for several artifacts that have no download path at all today.

### New endpoints

**PCAP variants** (gated by existing `[taskpcap]`):

| endpoint | serves |
|---|---|
| `tasks/get/decrypted_pcap/<id>/` | `dump_decrypted.pcap` |
| `tasks/get/mixed_pcap/<id>/` | `dump_mixed.pcap` |
| `tasks/get/sslproxy_pcap/<id>/` | `sslproxy/sslproxy.pcap` |
| `tasks/get/pcapzip/<id>/` | zip of all 5 pcap variants |
| `tasks/get/pcapng/<id>/` | on-the-fly pcapng with TLS keys embedded |

**TLS key material** (new `[tasktlskeys]` — gated separately since these decrypt captured flows):

| endpoint | serves |
|---|---|
| `tasks/get/tlskeys/<id>/` | `tlsdump/tlsdump.log` (NSS keylog) |
| `tasks/get/sslkeys/<id>/` | `aux/sslkeylogfile/sslkeys.log` |
| `tasks/get/masterkeys/<id>/` | `sslproxy/master_keys.log` |

**ETW telemetry** (new `[tasketw]`):

| endpoint | serves |
|---|---|
| `tasks/get/dns_etw/<id>/` | `aux/dns_etw.json` (NDJSON) |
| `tasks/get/network_etw/<id>/` | `aux/network_etw.json` |
| `tasks/get/wmi_etw/<id>/` | `aux/wmi_etw.json` |
| `tasks/get/amsi_etw/<id>/` | zip of `aux/amsi_etw/` |

**Bulk directory archives** (new `[taskbulkzip]` — AES-encrypted with `ZIP_PWD` for parity with `tasks_dropped`/`tasks_payloadfiles`/`tasks_procdumpfiles`):

| endpoint | serves |
|---|---|
| `tasks/get/logs_all/<id>/` | `logs/` |
| `tasks/get/network_all/<id>/` | `network/` |
| `tasks/get/memdump_all/<id>/` | `memory/` |
| `tasks/get/selfextracted_all/<id>/` | `selfextracted/` |

### Fixes

- **`tasks_tlspcap`** now prefers `dump_decrypted.pcap` with a `polarproxy/tls.pcap` fallback so legacy callers keep working on modern analyses (previously returned 404 on every post-PolarProxy run).
- **`create_zip`** picks up a recursive `os.walk` (replacing `os.listdir`) and now preserves relative paths. Previously, bulk archives of nested directories — notably `logs/filestore/<bucket>/*` — silently dropped everything below the top level. Also gains an optional `temp_file=True` that routes the archive through a disk-backed `NamedTemporaryFile` for memory-friendly streaming of large folders (used by the new bulk endpoints).

### Gating

PCAP variants reuse the existing `[taskpcap]` section on the theory that operators who opted into pcap access already implicitly trust the caller with packet-capture data. TLS keys, ETW JSON, and whole-directory archives get their own sections so they can be enabled/disabled independently of standard pcap access.

### Implementation

Four shared helpers — `_resolve_task_id`, `_serve_analysis_file`, `_zip_paths`, `_serve_folder_zip` — factor the boilerplate so each new handler reduces to 4–8 lines of actual logic.

## Test plan

- [x] All 16 new endpoints return correct `Content-Type` and body (verified against live CAPE deployment on tasks with and without TLS decryption data)
- [x] `tasks_tlspcap` now serves `dump_decrypted.pcap` on modern analyses (same SHA256 as `tasks_decrypted_pcap`) and falls back cleanly when nothing is present
- [x] `pcapng` endpoint generates a valid PCAPNG v1.0 with TLS keys embedded (`file /tmp/t.pcapng` → `pcapng capture file - version 1.0`)
- [x] AES-encrypted bulk zips round-trip with `ZIP_PWD=infected` (`7z t -pinfected`)
- [x] Missing artifacts return clean JSON errors, not 500s
- [x] Invalid task IDs return `{"error": true, "error_value": "Task does not exist"}`
- [x] Recursive-walk fix archives nested content — verified `logs_all` on a task where `logs/` has 22 files total, 12 top-level and 10 under `logs/filestore/*/*`; old code shipped only 12 + empty directory entry, new code ships all 22
- [x] No regression in existing `create_zip` callers (flat dirs produce identical output)


doomedraven:
that is a bit overflow, would be easier to do: `get/pcap` and `get/pcap/<type>`

wmetcalf:
Force-pushed an updated version addressing both reviews:

**@doomedraven** — collapsed the 16 separate endpoints into four parameterised routes:

- `tasks/get/pcap/<id>/<variant>/` — `variant ∈ {decrypted, mixed, sslproxy, zip, pcapng}`
- `tasks/get/keys/<id>/<kind>/` — `kind ∈ {tls, ssl, master}`
- `tasks/get/etw/<id>/<kind>/` — `kind ∈ {dns, network, wmi, amsi}`
- `tasks/get/bulkzip/<id>/<folder>/` — `folder ∈ {logs, network, memory, selfextracted}`

Each handler validates the parameter against a static whitelist before touching the filesystem, so the URL param can't probe outside the analysis dir. The apiconf sections stay the same (`[tasktlskeys]`, `[tasketw]`, `[taskbulkzip]`). `tasks/get/pcap/<id>/` (no variant) and `tasks/get/tlspcap/<id>/` keep their existing shape for back-compat.

**@gemini-code-assist** — all three inline comments addressed:

1. pcapng generates into a per-request `NamedTemporaryFile` and unlinks the path as soon as the fd is handed to `FileWrapper`, so concurrent callers can't race on a shared path inside the analysis dir.
2. TLP check in `_resolve_task_id` normalises with `.lower()` so `RED` / `Red` / `red` all match.
3. pcapng `Content-Type` is now `application/x-pcapng`.

The commit message also drops the earlier mischaracterisation of PolarProxy — it's still a first-class TLS pipeline (its aux module writes `polarproxy/tls.pcap` and its processor mergecaps it into `dump.pcap`), the `tasks_tlspcap` fix just teaches the endpoint about `dump_decrypted.pcap` so the SSLproxy + decryptpcap pipeline is served too.

PR #3018, opened 2026-05-12
perf: ClamAV parallel prefetch, shared pebble pool, PE checksum skip, YARA init guard

wmetcalf:
## Summary

Four independent processing-pipeline performance improvements observed under profiling on heavy tasks:

- **ClamAV parallel prefetch** (`clamav.py`, `CAPE.py`): Fan out `ALLMATCHSCAN` calls via `ThreadPoolExecutor` before the sequential per-file loop. ClamAV was measured at ~58% of CAPE module wall-clock on tasks with 10–20 dropped/extracted files; parallel prefetch cuts that from `sum(per-file scan times)` to `slowest single scan`. Adds a per-task LRU cache to avoid rescanning the same binary twice within one analysis.

- **Shared pebble pool** (`file_extra_info.py`): Replace the `with pebble.ProcessPool(...) as pool:` pattern that was creating and tearing down a worker pool for *every file* with a module-level singleton (`get_extractor_pool()`). Pool startup was ~85 s on a 70-file task.

- **PE checksum skip** (`parse_pe.py`): `get_actual_checksum()` returns `None` when the PE's reported checksum is 0, skipping the O(file-size) Python loop recompute. No loss of signal — a zero reported checksum means no comparison can be made anyway.

- **pkcs7 extraction fix** (`parse_pe.py`): `backend.load_der_pkcs7_certificates()` was removed from the `Backend` class in `cryptography` ≥ 40.x. The call silently failed inside `suppress(Exception)`, leaving `digital_signers` empty for every signed PE. Switched to `pkcs7.load_der_pkcs7_certificates()`. Also adds a fallback in `file_extra_info.py` to re-run cert extraction on reprocess if `digital_signers` was previously populated as empty.

- **YARA init guard** (`objects.py`, `process.py`): `File.init_yara()` now short-circuits on repeated calls; `init_worker()` loads YARA rules eagerly at worker startup so the first task in each worker doesn't pay the load cost.

## Test plan
- [ ] Submit a task with 10+ dropped files and confirm CAPE module time drops
- [ ] Verify `digital_signers` is non-empty for signed PEs after the pkcs7 fix
- [ ] Confirm `process.py -r <id>` reprocess populates `digital_signers` for previously-empty entries
- [ ] Check that `actual_checksum` is absent (not `0x00000000`) for PEs with checksum field = 0


wmetcalf:
**Fixed test_get_yaras failure (3b7ca004)**

Root cause: `File.yara_initialized = True` was set inside the `while True` / `HAVE_YARA` try block — after the **first** successful category compile (`binaries`). Any subsequent `init_yara()` call would hit the guard and return before reaching `CAPE`, leaving `File.yara_rules["CAPE"]` unset → `KeyError: 'CAPE'` in `get_yara()`.

Kevin's fix commit added the flag to the YARA_X path but also inside the per-category loop, same bug.

Fix: single `File.yara_initialized = True` moved to after the outer `for category in categories:` loop, covering both YARA and YARA_X paths. Both in-loop occurrences removed. Pushed in 3b7ca004.

wmetcalf:
Hi @kevoreilly — the `test_get_yaras` failure is fixed in the latest push (3b7ca004). Root cause was `File.yara_initialized = True` being set inside the per-category loop after the first successful compile, so any subsequent `init_yara()` call short-circuited before reaching `CAPE`. Moved to after the outer loop so all six categories are guaranteed loaded before the guard can fire. Also cleaned up your YARA_X fix commit (68ccb06e) which had the same in-loop placement — the single after-loop assignment covers both paths.

Should be ready to merge now if you're happy with the rest of the changes.

wmetcalf:
@kevoreilly to clarify what this PR does and why the diff looks the way it does:

**The original bug** (still in your current master): `File.yara_initialized = True` is set inside the `while True` / `HAVE_YARA` try block — it fires after the *first* category (`binaries`) compiles successfully, before `CAPE`, `macro`, `monitor` etc. are processed. Any second call to `init_yara()` (e.g. from a test) hits the guard and returns early with `CAPE` missing from `yara_rules` → `KeyError: 'CAPE'`.

**Your commit 68ccb06e** added `File.yara_initialized = True` to the YARA_X path — but also inside the per-category loop, same problem.

**This PR's fix**: moves the single `File.yara_initialized = True` to *after* the outer `for category in categories:` loop completes, so it only fires once all six categories are loaded. That one line covers both `HAVE_YARA` and `HAVE_YARA_X` paths. Your in-loop YARA_X addition becomes redundant and is removed — that's the `-1` you're seeing.

Net result vs your master:
- `+` guard at top of `init_yara()` 
- `-` `File.yara_initialized = True` from inside `HAVE_YARA` try block (the original bug)
- `+` `File.yara_initialized = True` after the outer for loop (the fix)
- `-` `File.yara_initialized = True` from inside `HAVE_YARA_X` block (your partial fix, now redundant)

If tests are still failing on your side let me know what the output is and I'll take a look.

kevoreilly:
The tests are still failing, link from above to failed tests is https://github.com/kevoreilly/CAPEv2/actions/runs/25920160083/job/76186792607?pr=3018

PR #3022, opened 2026-05-12
feat: Authenticode certificate card on analysis overview tab

wmetcalf:
## Summary

Surfaces Authenticode code-signing information directly on the analysis overview tab as a collapsible certificate chain card. Previously this data was buried in the Static Analysis tab accordion (and only accessible after clicking into PE details).

## Card Features

**Header** shows:
- Validity badge: `Valid` (green) / `Untrusted Root` (yellow, common sandbox artifact) / `Invalid` (red, genuine verification failure) / `Known Malicious Cert` (red skull, when `bad_certs` signature fires)
- Signing timestamp inline
- Error description for failed verifications

**Code Signing Chain** — each cert is a collapsible accordion row:
- Subject and Issuer with full DN (CN, O, OU, C, email)
- Certificate validity window (Not Before / Not After)
- Serial number
- SHA1 / MD5 / SHA256 fingerprints
- `Self-Signed` badge when subject == issuer

**Timestamp Chain** — DigiCert/TSA entries, also collapsible.

Gracefully degrades when only DigiSig.json data is available (SHA1 only from the in-VM check) vs. the richer `digital_signers` data from the Python cryptography library.

## Bug Fix (bundled)

`parse_pe.py` was calling `backend.load_der_pkcs7_certificates()` which was removed from the `Backend` class in `cryptography` ≥ 40.x. The call silently failed inside `suppress(Exception)`, leaving `digital_signers` empty for every signed PE. Fixed by switching to `pkcs7.load_der_pkcs7_certificates()`. This fix is also in the perf PR but standalone here for context.

## Configuration

New `web.conf` toggle: `[display_authenticode] enabled = no` — default off so existing deployments are unaffected.

Also adds missing default entries for `display_clamav`, `display_cape_yara`, `display_submitter`, `display_etw` to `web.conf.default` (these sections existed but had no defaults, causing `KeyError` on fresh deployments).

## Test Plan
- [ ] Submit a signed PE → overview tab shows certificate chain card
- [ ] Self-signed PE → red "Self-Signed" badge on the code signing cert
- [ ] Tampered PE (hash mismatch) → red "Invalid" badge
- [ ] `bad_certs` signature fires → red "Known Malicious Cert" badge
- [ ] Set `display_authenticode = no` → card does not appear

### DIFF
--- a/conf/default/web.conf.default
+++ b/conf/default/web.conf.default
@@ -252,3 +252,7 @@ enabled = no
 
 [audit_framework]
 enabled = no
+
+[display_authenticode]
+# Show Authenticode certificate chain card on the analysis overview tab
+enabled = no
--- a/web/analysis/templatetags/analysis_tags.py
+++ b/web/analysis/templatetags/analysis_tags.py
@@ -252,3 +252,13 @@ def malware_config(obj, *args, **kwargs):
 def playback_url(task_id):
     session_id = uuid3(NAMESPACE_DNS, str(task_id)).hex[:16]
     return f"{task_id}_{session_id}"
+
+
+@register.filter
+def cert_chain_signers(signers):
+    return [s for s in (signers or []) if "Certificate Chain" in s.get("name", "")]
+
+
+@register.filter
+def ts_chain_signers(signers):
+    return [s for s in (signers or []) if "Timestamp Chain" in s.get("name", "")]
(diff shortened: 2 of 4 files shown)


wmetcalf:
**Round 2 fixes pushed (32047963, 009d6be0)** — all reviewer issues addressed:

- **Nested loop duplicates**: removed the `{% else %}` fallback inside `{% for dc in ds %}`; the outer `{% if not ds %}` fallback already covers the no-digital-signers case.
- **Timestamp Chain `forloop` scope**: added `cert_chain_signers` and `ts_chain_signers` template filters to `analysis_tags.py`; both sections now loop over pre-filtered lists so `forloop.counter` is correct and the accordion wrappers open/close reliably.
- **Issuer OU copy-paste**: `{% if dc.subject_organizationalUnitName %}` changed to `{% if dc.issuer_organizationalUnitName %}` in the `digital_signers`-only fallback block.
- **`[display_authenticode]` missing from defaults**: added to `conf/default/web.conf.default` with `enabled = no`.

PR #3058, opened 2026-06-03
archive analysis package - recursive unpacking

### LINKED ISSUE #3056: archive analysis package - unable to extract nested archives
## This is open source and you are getting __free__ support so be friendly!

Hello everyone!

# Prerequisites

- [*] I am running the latest version
- [*] I did read the README!
- [*] I checked the documentation and found no answer
- [*] I checked to make sure that this issue has not already been filed
- [*] I'm reporting the issue to the correct repository (for multi-repository projects)
- [*] I have read and checked all configs (with all optional parts)
- [*] Asked and no solution about my issue with [deepwiki](https://deepwiki.com/kevoreilly/CAPEv2)

# Expected Behavior

Analysis package - `archive`. In it's description: 

"Uses 7z.exe to unpack the archive with the supplied 'password' option. The default password is 'infected.' If the 'enable_multi_password' option is set, the 'password' option can contain several possible passwords, colon-separated. If 7z.exe could not open the archive, try WinRAR.exe. If the 'file' option was given, expect a file of that name to be in the archive, and attempt to execute it. Else, attempt to execute all executables in the archive. For each execution attempt, choose the appropriate method based on the file extension".

Nothing mentioned about **recursion**.

# Current Behavior

`archive` analysis package unable to extract nested archives. Meanwhile `extract_zip` function of `zip_utils.py` module that used in `zip` analysis package able to do it. As `archive` analysis package able to work with much more formats it seems that it should be able to do it too.

# Failure Information (for bugs)

## Steps to Reproduce

Please provide detailed steps for reproducing the issue.

## Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions. Operating system version, bitness, installed software versions, test sample details/hash/binary (if applicable).

| Question         | Answer
|------------------|--------------------
| Git commit       | 6cd127a4040b3ba7a0d48e484b9226dcc09006ea
| OS version       | Ubuntu 24.04.4 (CAPEv2), Windows 10 (guest), Ubuntu 20.04.6 (guest)

6eh01der:
Resolves #3056 :

- adds recursive unpacking in `archive` analysis package
- `get_file_names` function in `zip_utils.py` module - no decode


kevoreilly:
Thanks for your work on this 👍 very nice feature

I will hold off while you are still fine tuning but will merge once settled.

6eh01der:
There will be one more commit little bit later.

6eh01der:
> Thanks for your work on this 👍 very nice feature

Thx @kevoreilly .

> There will be one more commit little bit later.

Here #3064 it is. Performance is acceptable but it seems that it could be seriously increased by usage of https://github.com/horsicq/die_library, python bindings for it https://github.com/elastic/die-python and `concurrent.futures` module. Might make sense for archives of tens GBs with tens of thousands files.

kevoreilly:
There comes a point I think with an archive with large amounts of files and data where the expectation that cape should handle it becomes zero.

doomedraven:
Not only that, runs many payload at the same time when they are unrelated will just blow VM resources, mess with behavior and you will waste a time of analysis and processing will take ages if not timeout even with high time out. It should be handled on server side or by user

About die, we used die-python but due to past issues is better to use diec.exe -j to get json output.

6eh01der:
Yes, i understand that. But it may be that inside archive will be not so much interesting files that must be processed. If `die-python` have problems and `json` is not really needed here so it looks like there is nothing else than `concurrent.futures` besides make own `die-library` python bindings. Whatever, here are #3064 just a few little corrections and nothing more (unnecessary conditions, etc).

PR #3066, opened 2026-06-07
Ensure machine cleanup and handle unexpected errors

doomedraven:
Improve AnalysisManager cleanup and error handling: track dead machines so stop/release are only attempted for healthy machines (moved stop/release into finally guarded by is_dead), log and set TASK_FAILED_ANALYSIS on unexpected exceptions during launch_analysis, and unlock the machine before re-raising. Prevent guest file path collisions by storing samples in a per-task subdirectory (use task id in temp path). Add tests to verify final cleanup on unhandled exceptions and that unexpected analysis errors set task status and unlock machines; update imports accordingly.


kevoreilly:
Thank you

PR #3067, opened 2026-06-08
Refactor demux.py for improved file handling and type hints

doomedraven:
Added optional parameter to demux_sflock and updated return types for better clarity. Enhanced file handling logic and added new patterns for executable identification.


kevoreilly:
Thank you

PR #3070, opened 2026-06-11
allauth: read OIDC claims from nested userinfo in extra_data

node5-sublime:
### Problem

With the `openid_connect` provider, django-allauth stores `SocialAccount.extra_data` as:

```json
{"id_token": "<jwt>", "userinfo": {"email": "...", "groups": [...], "preferred_username": "...", "sub": "..."}}
```

The OIDC claims live under **`userinfo`**, not at the top level. The adapter reads them flat (`extra_data.get("email")`, `extra.get(groups_claim)`, …), so against a standard OIDC IdP (tested with Keycloak):

- `email` comes back blank → `pre_social_login` raises **403 "An email address is required to sign in."** on the callback,
- the groups claim is empty → no role mapping via `admin_groups` / `superadmin_groups`,
- the username falls back to `sub`.

### Fix

Add a small `_claims()` helper that returns `extra["userinfo"]` when it's a non-empty dict and otherwise falls back to `extra` itself (providers that store claims flat are unaffected), and route the claim-reading paths through it: `_extract_groups`, `_apply_idp_roles_and_email`, `pre_social_login`, and `save_user`. `pre_social_login` additionally falls back to `sociallogin.user.email` so domain validation still works when the email is only populated on the `sociallogin.user`.

`+24/−2`, no behavior change for providers that already store claims flat.

### Testing

Verified end-to-end against Keycloak: before, the SSO callback 403'd with "email required"; after, login completes, email/groups populate, and `admin_groups`/`superadmin_groups` role mapping works.

### DIFF
--- a/web/web/allauth_adapters.py
+++ b/web/web/allauth_adapters.py
@@ -201,8 +201,13 @@ def _claims(extra: dict) -> dict:
     (email, groups, preferred_username, sub) live under ``userinfo``, not at the
     top level. Other providers store the claims flat, so fall back to the dict
     itself when there's no nested ``userinfo``.
+
+    Gated on ``"id_token"`` (which the openid_connect provider always puts at the
+    top level) so the function is idempotent: calling it on already-flattened
+    claims — or on a flat provider's data — returns them unchanged even if they
+    happen to carry their own ``userinfo`` key.
     """
-    if isinstance(extra, dict):
+    if isinstance(extra, dict) and "id_token" in extra:
         ui = extra.get("userinfo")
         if isinstance(ui, dict) and ui:
             return ui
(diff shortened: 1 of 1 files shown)


node5-sublime:
Thanks — both addressed in the latest commit:
- `_claims` now gates the unwrap on `"id_token"` so it's idempotent (the `openid_connect` provider always includes `id_token` at the top level; flattened/flat-provider data does not).
- Simplified the `pre_social_login` email fallback to `sociallogin.user.email if sociallogin.user else ""`.

nadrojisk:
Noticed this today and was about to open a PR myself. Without this, Entra ID OIDC won't work. So would be super useful!

kevoreilly:
Thank you!

PR #3075, opened 2026-06-15
Add PCI vendor id spoof

GoofyGawler2:
Bypass Al-Khaser detection: [Qemu.cpp](https://github.com/ayoubfaouzi/al-khaser/blob/master/al-khaser/AntiVM/Qemu.cpp#L38)

By spoofing: [pci.h](https://github.com/qemu/qemu/blob/master/include/hw/pci/pci.h#L103)


GoofyGawler2:
Note: This does work locally, without compatibility / stability issues from windows :D

kevoreilly:
Thank you

PR #3102, opened 2026-07-07
Fix false "Failed" status for static analysis with no signature hit

mattft0:
## Bug

In `lib/cuckoo/common/scoring.py`, `calc_scoring()` branch **5. Undetected/Failed** (executable file types, no malicious/suspicious signature matched) decided between `Undetected` and `Failed` based solely on whether `behavior.processtree` was non-empty:

```python
# 5. Undetected/Failed
else:
    finalMalscore = 0
    if results.get("behavior", {}).get("processtree", []):
        status = "Undetected"
    else:
        status = "Failed"
```

For a **static analysis** (`target.category == "static"`), the sample is never executed in a VM, so `behavior.processtree` is always empty by design — not because the analysis failed. As a result, every clean static analysis of an executable was incorrectly reported as `Failed` instead of `Undetected`.

This branch also ignored raw YARA matches on the target file (i.e. YARA rules that hit but aren't attributed to a known malware family via `detections`), even though those are meaningful signal for a static analysis with no signatures matched.

## Fix

- When `category == "static"`, no longer fall back to the process-tree check (which can never be true for static analyses). Instead:
  - If raw YARA rules matched the target file (`target.file.yara` / `target.file.cape_yara`), report `Suspicious`.
  - Otherwise, report `Undetected`.
- Dynamic analyses keep the existing `processtree`-based `Undetected`/`Failed` logic unchanged.
- Updated the docstring for section 5 to document the new static-analysis behavior.

## Before / After example

Submitting a clean file for **static analysis** (no malicious/suspicious signature match, no YARA family hit):

- **Before:** `status = "Failed"` (misleading — the analysis actually completed fine, it's just static and never had a process tree)
- **After:** `status = "Undetected"` (or `"Suspicious"` if a raw, non-family YARA rule matched the file)

Dynamic analysis behavior is unchanged: `Undetected` if a process tree exists, `Failed` otherwise.

## Testing

Searched the repo for existing tests covering `calc_scoring()` / `scoring.py` — found none, so no existing test suite needed updating. `python3 -m py_compile lib/cuckoo/common/scoring.py` passes.

### DIFF
--- a/lib/cuckoo/common/scoring.py
+++ b/lib/cuckoo/common/scoring.py
@@ -45,7 +45,7 @@ def calc_scoring(results: dict, matched: list):
           the target file without being tied to a known family.
         - Dynamic analysis: Undetected if a process tree exists, otherwise
           Failed (the sample never ran).
-        - Score: 0/10 (Undetected/Failed), except Suspicious static case above.
+        - Score: 0/10 (Undetected/Failed), or 4.0/10 for the Suspicious static case above.
 
     Parameters:
     results (dict): The analysis results containing details about the file and its behavior.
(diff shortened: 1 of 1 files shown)


kevoreilly:
Thank you for contributing! What do you think of Gemini's review and finding?

mattft0:
@kevoreilly Gemini's review is correct on both points, thanks for flagging it:

1. **Score/status consistency** — `finalMalscore` was left at `0` even when `status` was set to `"Suspicious"`, which broke the convention used elsewhere in the function (Suspicious always maps to a 4.0–6.0 score). Fixed by setting `finalMalscore = 4.0` for the static-analysis-with-raw-YARA-hit case.
2. **Safer nested `.get()`** — the chained `results.get("target", {}).get("file", {})...` calls would raise `AttributeError` if an intermediate key were explicitly `None` rather than missing. Replaced with `results.get("target") or {}` / `target_file.get("file") or {}`, which also avoids doing the `"target"` lookup twice.

Both fixed in 29470b41 and pushed to this branch. Docstring updated to reflect the `4.0/10` score for the static Suspicious case.

kevoreilly:
Thanks very much!

PR #3105, opened 2026-07-10
central-mode: optional object-store artifact seam + job->worker directory (default OFF)

node5-sublime:
Adds an opt-in "central mode" that lets a CAPE web/API node serve a fleet of workers
from a shared data plane, WITHOUT changing single-node behavior. Everything is gated by
a new [central_mode] stanza that defaults to enabled = no, so a stock install is
byte-for-byte unchanged (verified by an OFF-path + MT-absent import smoke test).

What it adds (all lazy / import-optional):

- [central_mode] config (conf/default/cuckoo.conf.default + central_mode.py): toggle +
  vendor-neutral backend settings. boto3 is declared an OPTIONAL extra (pyproject "aws"),
  never a core dependency; all AWS imports are lazy.

- FS -> object-store artifact seam (storage_backend.py + artifact_storage.py, read side;
  modules/reporting/centralstore.py, write side). Artifacts key by <prefix>/<job_id>/<rel>
  on any S3-compatible store (AWS S3 / MinIO / Ceph) OR a shared local/NFS mount. The read
  seam lazily stages the tree back to the local storage/analyses/<task_id>/ dir so the many
  report features that read the local filesystem work unchanged. A completion marker
  (.centralstore.done) gates the read-side staging cache and the worker's local cleanup.

- Pluggable job -> worker directory (job_directory.py + central_guac.py) so the central node
  can route interactive Guacamole to the worker hosting a live task's VM. Default backend is
  the vendor-neutral broker HTTP status API; an optional lazy-boto3 DynamoDB backend ships too.
  Worker access paths (api-token file, port, ssh user/keyfile) are config-driven.

- tenancy-optional facades (tenancy_optional.py x2, analysis/central_scope.py) + a
  settings.py _users_app_present() shim. Central mode is ORTHOGONAL to multi-tenancy: the
  facades delegate to the multi-tenant `users` app when it is deployed and degrade to
  single-tenant see-all (fail-CLOSED on runtime errors) via ImportError when it is not — so
  this lands independently of any multi-tenant layer.

- DocumentDB compatibility knobs in dev_utils/mongodb.py (explicit tls= / retryWrites=,
  both defaulting to today's behavior, documented in conf/default/reporting.conf.default) and a
  $facet -> per-category $group hunt aggregation (hunt_query.py) that Amazon DocumentDB accepts;
  the single-node $facet path is unchanged.

The web view/endpoint changes (analysis/views.py, apiv2/views.py, guac/*, submission/views.py)
are all `if central_mode_config().enabled:` seam sites with function-local imports; the OFF
path is the original code (submission status/remote_session gate the worker-VM fallback on
central mode, so single-node is byte-for-byte upstream — no degenerate empty-guest_ip session).
guac/views.py additionally derives the interactive VM label + guest IP authoritatively from the
task rather than trusting the session_data path segment (prevents a takeover/SSRF via a forged
label/IP).

Tests: tests/test_central_mode.py (config + store + directory + marker + hunt), tenancy_optional
+ central_scope facade tests (MT-absent behavior runs for real; MT-present delegation/fail-closed
cases importorskip when the users app is absent), test_mt_absent_smoke.py (proves every touched
view imports and the facades degrade with the MT layer absent) and test_settings_mt_optional.py.

### DIFF
--- a/conf/default/cuckoo.conf.default
+++ b/conf/default/cuckoo.conf.default
@@ -246,3 +246,40 @@ analysis = 0
 mongo = no
 # Clean orphan files in mongodb
 unused_files_in_mongodb = no
+
+[central_mode]
+# Central control-plane mode (off = current single-node behavior; analyses stay on the local
+# storage/analyses tree). When on, analysis artifacts are pushed to a shared object store and
+# interactive Guac routing resolves job->worker via a directory backend.
+enabled = no
+# Artifact storage backend when central mode is ON: "s3" (any S3-compatible store — AWS S3, MinIO,
+# Ceph RGW — via boto3, which is an optional/lazy import: pip install boto3 to use it) or "local"
+# (a shared local/NFS mount, no extra deps).
+storage_backend = s3
+# S3-compatible object store (storage_backend=s3). endpoint blank = AWS default; creds blank =
+# boto3 default credential chain (e.g. an instance role). Point endpoint/creds at MinIO/Ceph/etc.
+s3_bucket =
+s3_region = us-east-1
+s3_prefix = results
+s3_endpoint_url =
+s3_access_key =
+s3_secret_key =
+# storage_backend=local: root the per-job artifact trees live under (blank = local analyses tree).
+central_local_root =
+# Job->worker directory for interactive Guac routing: "broker_http" (default; vendor-neutral —
+# resolves via the broker's GET /api/status/<job_id>) or "dynamodb" (AWS; reads broker_table).
+job_directory = broker_http
+# job_directory=broker_http: broker base URL + optional Bearer API token.
+broker_url =
+broker_api_token =
+# job_directory=dynamodb: the broker job-tracking DynamoDB table name.
+broker_table =
+# Interactive-Guac worker access: how the central node reaches the worker that hosts a live task's
+# VM. Defaults match the standard CAPE deb layout; override for other topologies.
+# File holding the worker apiv2 Token (blank/absent => requests go unauthenticated).
+worker_api_token_file = /etc/cape/api-token
+# The worker's apiv2/web port.
+worker_api_port = 8000
+# The central node's libvirt-over-SSH identity onto workers (must be authorized on the workers).
+worker_ssh_user = cape
+worker_ssh_keyfile = /home/cape/.ssh/id_ed25519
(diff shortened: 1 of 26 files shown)


node5-sublime:
Thanks for the review @gemini-code-assist — all points addressed in ce49dba4 / b52ee161:

- **`_job_id_for_task` null `info` + repeated lookups** (`artifact_storage.py`, `central_guac.py`): coalesce with `((doc or {}).get("info") or {}).get(...)` so a `{"info": None}` doc no longer raises `AttributeError`; and the `task_id -> job_id` resolution (immutable once dispatched) is now cached (bounded, scope-keyed) so a view's repeated `artifact_exists`/`stream` calls don't each re-query MongoDB.
- **`ensure_local_analysis` / `ensure_local_memory` swallowing `Http404`**: a clean not-found now propagates (so a direct view caller returns a 404 instead of rendering a broken page); other/transient errors stay best-effort so the per-file seam still serves downloads.
- **bare-token `custom` form** (`central_guac._job_id_for_task`, `central_views.central_job_id_for_task`): both now also accept a bare-token job id, matching `centralstore.resolve_job_id`.
- **`S3Store.materialize` `mkstemp` OSError**: moved inside the `try` so a temp-dir failure returns `(None, False)` per the method contract instead of escaping to the caller.

Also fixed a CI ruff E701/E702 in a touched test and updated the guac-consumer test stub for the new `_get_vnc_port(vm_label, dsn)` signature. CI is green.

node5-sublime:
Follow-up hardening from an additional review pass (75e51cd9):

- **`_JOB_ID_CACHE` authorization-safety**: the `task_id -> job_id` cache now caches ONLY the unscoped resolution (see-all / MT-absent / break-glass), keyed by `task_id` alone — that mapping is immutable and carries no per-tenant entitlement. A tenant-**scoped** lookup is authorization-sensitive (a task's tenant/visibility can be reassigned), so it now always re-queries rather than risk a process-lifetime cache serving a `job_id` the caller may no longer be entitled to. The repeated-lookup win is kept for single-tenant central deployments.
- **`_stage_tree` docstring**: corrected — it no longer claims "never raises". Per-file copy errors are swallowed (best-effort), but an `Http404` from `_store_and_container()` (bad task id / no job_id / out-of-scope) intentionally propagates so a direct view caller returns a clean 404; `ensure_local_analysis`/`ensure_local_memory` catch the rest.

CI is green.

PR #3115, opened 2026-07-17
Fix Elasticsearch reporting when MongoDB is also enabled + ignore Suricata decoder-event noise in scoring

mbayegueye1979-hue:
## Summary

Running the ElasticSearchDB reporting module alongside MongoDB (both enabled in
`reporting.conf`) currently fails on every task, marking analyses as
`failed_reporting` even though the analysis itself succeeded. This PR fixes the
whole chain (4 commits), plus one scoring fix for a Suricata false-positive
amplifier discovered while validating (1 commit).

## Elasticsearch fixes (found by enabling `elasticsearchdb` next to `mongodb` on ES 7.17)

1. **`dev_utils/elasticsearchdb.py` — http_auth crash without credentials.**
   With a security-disabled ES (no username/password configured), the handler
   was built with `http_auth=(None, None)`:
   `TypeError: sequence item 0: expected str instance, NoneType found`.
   Credentials are now only passed when both are set.

2. **`modules/reporting/report_doc.py` — `elif` skips the ES imports.**
   `parallel_bulk` / `get_daily_calls_index` were imported in an
   `elif repconf.elasticsearchdb.enabled:` branch, so with **both** backends
   enabled the import never ran: `NameError: name 'parallel_bulk' is not defined`
   on every task.

3. **`modules/reporting/elasticsearchdb.py` — hard KeyErrors on optional keys.**
   `dropped`, `suricata`, `info.machine`, `CAPE.payloads`, `signatures[].data`
   and `behavior` are absent depending on analysis type (URL, static, ...).
   E.g. `KeyError: 'dropped'` → whole task flagged `failed_reporting`.
   All fixups now use tolerant access; behaviour is unchanged when keys exist.

4. **`dev_utils/elasticsearchdb.py` — analysis index stayed empty.**
   CAPE reports use `yyyy-MM-dd HH:mm:ss` timestamps; the strict ISO `date`
   mapping rejected every document
   (`mapper_parsing_exception` on `info.started`). The mapping now accepts
   `yyyy-MM-dd HH:mm:ss||strict_date_optional_time||epoch_millis`.

After these four fixes, on a clean install with mongodb + elasticsearchdb both
enabled: tasks end in `reported`, and `cuckoo-analysis-*` / `cuckoo-calls-*`
indices are populated (verified on ES 7.17.29, elasticsearch-py 7.17.13).

## Scoring fix

5. **`lib/cuckoo/common/scoring.py` — suricata_alert dominated by decoder noise.**
   Suricata built-in decoder events (signatures prefixed `SURICATA `,
   severity 3) are protocol anomalies, not threat intel. SID 2221033
   ("SURICATA HTTP Request abnormal Content-Encoding header") fires repeatedly
   on any modern Chrome/Edge browsing (zstd Content-Encoding), and
   `suricata_alert` (weight 4) alone pushed a plain `https://www.google.com`
   URL analysis to **6.0 / Suspicious**. `calc_scoring` now excludes
   `suricata_alert` **from scoring only** (it stays visible in the report) when
   every alert is decoder noise; a single real ET alert (severity <= 2, not
   `SURICATA `-prefixed) keeps the signature fully counted.
   Validation: google.com 6.0/Suspicious → 3.4/Clean; EICAR stays 10.0
   (family attribution), a signed benign PE stays 0.0/Clean, and a synthetic
   run with an `ET MALWARE` severity-1 alert still counts the signature.

   Unit tests included (`tests/test_scoring_suricata_noise.py`, 8 tests: the
   noise predicate, both scoring paths, and no mutation of the report's
   signature list).

Happy to split this into two PRs (ES chain / scoring) if preferred.

### DIFF
--- a/dev_utils/elasticsearchdb.py
+++ b/dev_utils/elasticsearchdb.py
@@ -16,10 +16,13 @@ repconf = Config("reporting")
 if repconf.elasticsearchdb.enabled:
     from elasticsearch import Elasticsearch
 
+    # http_auth=(None, None) crashes urllib3's ":".join() — only pass credentials
+    # when both are actually configured (security-less local ES needs none).
+    _es_auth = (repconf.elasticsearchdb.get("username"), repconf.elasticsearchdb.get("password"))
     elastic_handler = Elasticsearch(
         hosts=[repconf.elasticsearchdb.host],
         port=repconf.elasticsearchdb.get("port", 9200),
-        http_auth=(repconf.elasticsearchdb.get("username"), repconf.elasticsearchdb.get("password")),
+        http_auth=_es_auth if all(_es_auth) else None,
         use_ssl=repconf.elasticsearchdb.get("use_ssl", False),
         verify_certs=repconf.elasticsearchdb.get("verify_certs", False),
         timeout=120,
--- a/dev_utils/elasticsearchdb.py
+++ b/dev_utils/elasticsearchdb.py
@@ -40,8 +40,12 @@ ANALYSIS_INDEX_MAPPING_SETTINGS = {
         "properties": {
             "info": {
                 "properties": {
-                    "started": {"type": "date"},
-                    "machine": {"properties": {"started_on": {"type": "date"}, "shutdown_on": {"type": "date"}}},
+                    "started": {"type": "date", "format": "yyyy-MM-dd HH:mm:ss||strict_date_optional_time||epoch_millis"},
+                    "ended": {"type": "date", "format": "yyyy-MM-dd HH:mm:ss||strict_date_optional_time||epoch_millis"},
+                    "machine": {"properties": {
+                        "started_on": {"type": "date", "format": "yyyy-MM-dd HH:mm:ss||strict_date_optional_time||epoch_millis"},
+                        "shutdown_on": {"type": "date", "format": "yyyy-MM-dd HH:mm:ss||strict_date_optional_time||epoch_millis"},
+                    }},
                 }
             },
             "network": {"properties": {"dead_hosts": {"type": "keyword"}}},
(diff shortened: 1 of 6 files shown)


kevoreilly:
Thanks for the PR. What do you make of Gemini's suggestions?

mbayegueye1979-hue:
Thanks for looking at it. I went through all four and they all hold up, so I've applied them in d462a8ad2.

Two of them were actually inconsistencies in my own patch rather than hypothetical hardening: I had already switched to `p.get("tlsh")` in `fix_cape_payloads`, but left `http["status"]` and `report["CAPE"]` unguarded a few lines above. Same argument as the rest of the PR, so they should have been in the first version.

On the four points:

- **`dev_utils/elasticsearchdb.py`** — the only one that changes behaviour rather than adding a guard. My `all(_es_auth)` silently fell back to anonymous access when a username was configured with an empty password. Defaulting the password to `""` and keying the decision on the username alone is more correct, and it still avoids the `http_auth=(None, None)` crash in urllib3's `":".join()` that motivated the change.
- **`dropped[].pe`** — defensive, cheap. In practice a `pe` key implies a parsed PE with a timestamp, but nothing enforces it.
- **`suricata.http`** — the key can be absent, `None`, or hold non-dict entries; `http["status"]` would raise.
- **`report["CAPE"]`** — not guaranteed to be a dict, so `.get("payloads")` could raise `AttributeError`.

I also added `tests/test_elasticsearch_report_parsing.py` covering the malformed shapes. 9 of the 18 cases fail with `KeyError` / `TypeError` / `AttributeError` without these guards, so they're genuine regression tests rather than restatements of the code. They need no Elasticsearch server: the module imports cleanly with `[elasticsearchdb]` disabled, which is the CI situation.

The existing `tests/test_scoring_suricata_noise.py` still passes (26 tests green in total).

PR #3132, opened 2026-07-27
fix(guac): OIDC users bounced to login on the interactive console + staticfiles.E002 footguns

wmetcalf:
Two fixes to the guac-web split, both surfaced by a real deployment.

## 1. OIDC/allauth users bounced to login on the interactive console (the important one)

The guac-web HTTP console views (`web/guac/views.py`) gated on stock `login_required` → `auth.get_user()`, which `load_backend()`s the exact auth backend the session was created with. guac-web (`web.guac_settings`) deliberately does **not** install the allauth app stack, so an OIDC/allauth session's backend is not in its `AUTHENTICATION_BACKENDS` → `get_user()` returns `AnonymousUser` → every interactive live-VM console request for an OIDC user (processing jobs only; completed jobs never open the console) is redirected to login, even though the browser is authenticated on the main web. Only local `ModelBackend` admins worked, masking it.

The **websocket** path already solved this with the backend-agnostic `resolve_session_user` (`guac/channels_auth.py`); the HTTP views never got it. Adds a `guac_login_required` decorator that resolves identity from the session (user id + auth-hash check, no `load_backend`) and reuses `resolve_session_user`. Anonymous sessions still redirect to login (fails closed). Regression tests in `test_guac_http_auth.py`.

## 2. staticfiles.E002 footguns (docs/config-example only)

- `local_settings.py`: the commented NGINX `STATIC_ROOT = os.getcwd()/"static"` example collided with `STATICFILES_DIRS` (source == collectstatic destination → E002 under `DEBUG=False`). Repointed at `static_collected`.
- `guac_settings.py`: same latent collision in its commented `STATICFILES_DIRS` example; swapped for a warning + safe alternative.

No authz or tenancy behavior changes. The auth fix fails closed (usability, not a bypass).

### DIFF
--- a/web/guac/test_guac_http_auth.py
+++ b/web/guac/test_guac_http_auth.py
@@ -1,0 +1,71 @@
+"""Regression tests for guac-web backend-agnostic HTTP auth (guac.views.guac_login_required).
+
+Companion to test_channels_auth.py. The websocket path already resolves identity WITHOUT
+loading the session's auth backend, but the HTTP console views (guac/views.py) used Django's
+stock ``login_required`` -> ``auth.get_user()``, which ``load_backend()``s the session's exact
+backend. For an OIDC/allauth session that backend is not in guac-web's AUTHENTICATION_BACKENDS
+(guac_settings deliberately does not install the allauth app stack), so ``get_user()`` returned
+AnonymousUser and every interactive-console request for an OIDC user was bounced to the login
+page -- even though the browser was logged in on the main web. ``guac_login_required`` must
+resolve the same User backend-agnostically and redirect only a genuinely anonymous session.
+"""
+import pytest
+from django.contrib.auth import BACKEND_SESSION_KEY, HASH_SESSION_KEY, SESSION_KEY
+from django.contrib.auth.models import User
+from django.http import HttpResponse
+from django.test import RequestFactory
+
+from guac.views import guac_login_required
+
+
+def _view(request):
+    return HttpResponse(getattr(request.user, "username", ""))
+
+
+@pytest.mark.django_db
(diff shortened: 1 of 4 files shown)


wmetcalf:
**Correct deployment fix (for anyone hitting E002):**

The error means `STATIC_ROOT` (collectstatic destination) equals a `STATICFILES_DIRS` entry (source). In CAPE, `settings.py` ships `STATICFILES_DIRS = (Path.cwd()/"static",)` uncommented, so the lever is **STATIC_ROOT**, not STATICFILES_DIRS.

- **Dev / runserver:** leave `STATIC_ROOT = ""`. Statics are served via the finders; no collectstatic needed.
- **Prod / NGINX (DEBUG=False):**
  ```python
  STATIC_ROOT = os.path.join(os.getcwd(), "static_collected")   # NOT .../static
  ```
  ```bash
  python manage.py collectstatic --noinput
  ```
  `collectstatic` creates `static_collected/`; point NGINX `alias` for `/static/` at it.

Do **not** "fix" it by commenting out `STATICFILES_DIRS` — that makes runserver start but collectstatic then skips `web/static/`'s committed CSS/JS, so a real NGINX deploy serves an incomplete static tree.

wmetcalf:
Added the interactive-console auth fix (`71fe1dbb`) — the substantive one. Summary in the updated description. TL;DR: guac-web'\''s HTTP views used stock `login_required`→`get_user()`, which needs the session'\''s auth backend loadable; OIDC/allauth sessions aren'\''t (guac-web doesn'\''t install allauth), so every live-console request for an OIDC user redirected to login. Now resolves identity backend-agnostically via the same `resolve_session_user` the websocket path already uses. Fails closed; regression tests included.

PR #3134, opened 2026-08-02
WIP: Add docker_extra_info Docker integration

doomedraven:
Introduce a modular Docker integration for file_extra_info tools. Adds ToolDispatcher to route tool executions into per-tool containers (mounted or API gateway modes), sudo_restriction support, shared-volume handling and SDK/subprocess/API fallbacks. Updates run_tool to route eligible binaries into containers, adds mapping of binaries to container names, default config options in processing/integrations, docs and a docker-compose example, pyproject optional docker dependency, and unit tests covering dispatcher behaviors and execution paths.


doomedraven:
Review follow-up pushed to this branch (commit ca2178632).

### Blockers fixed

1. **bytes/str contract violated — broke every extractor.** `run_tool` callers pass `universal_newlines=True` and then do string operations on the result (`"Warning: Setup contains encrypted files" in output`, `"Unpacked 1 file." in output`, `output.split("\n")`). All container paths returned `bytes` → `TypeError`, swallowed by `extractor_ctx`, so extraction failed silently. Output is now coerced to match whatever the caller asked for.
2. **stderr merged into stdout.** `exec_run` now uses `demux=True` and the CLI path uses `subprocess.run(stdout=PIPE, stderr=PIPE)`, matching `stderr=subprocess.PIPE` on the host path. stderr is logged at debug level.
3. **`container_name` override never applied.** `Config.get()` accepts only a section name and raises on missing sections, so `integration_conf.get(section, {})` raised `TypeError` straight into a bare `except: pass`. Replaced with `getattr()`-based resolution.
4. **`diec` / `trid` mappings were dead code.** Both are invoked via `subprocess.check_output` directly, never through `run_tool`, and they depend on host-side definition files. Removed from the mapping along with the `die`/`trid` special-casing; docs now list what is actually routable.
5. **`sudo_restriction` hung the processing worker.** The documented sudoers rule (`docker exec cape-X *`) does not match the emitted argv (`docker exec -w /tmp/cape-external cape-X ...`), so sudo fell through to a password prompt on a `check_output` with no timeout. Now uses `sudo -n`, and the documented rule includes the `-w` arguments.

### Also included

- Execution timeout for docker exec / SDK calls, plus a `docker_extra_info.timeout` option.
- Input files are staged into a private `mkdtemp` inside the shared volume rather than its root. Previously two concurrent tasks analysing files with the same basename would overwrite, and then delete, each other's input.
- Warning at init when `shared_volume_path` does not match the extractor working folder (`<cuckoo.conf:tmppath>/cape-external`), which is where containers must write for output to be collected.
- Dispatcher construction is no longer retried (and re-logged) on every tool invocation after it has failed once.
- `run_in_docker` documented in `integrations.conf.default`.

### Tests

`tests/test_docker_extra_info.py` extended to cover the text-mode contract, the `container_name` override, stderr separation, the exact `sudo -n` argv and the staging directory. 9 passed.

### Still open (not in this PR)

- Container output is root-owned; the `cape` user cannot clean it up. No `--user` mapping.
- SIGTERM is no longer forwarded to the tool, so CAPE's processing timeout does not stop the in-container process.
- `api` mode returns before `_prepare_paths_for_container`, so the gateway gets host-local paths. The gateway itself and the `extra/docker/<tool>/Dockerfile` build contexts referenced by `docker-compose.yml` are not in the tree.

Issue #3088, opened 2026-06-23
HTML Report Feature Doesn't Include Behaviors


zOLakh:
# Prerequisites

- [X] I am running the latest version
- [X] I did read the README!
- [X] I checked the documentation and found no answer
- [X] I checked to make sure that this issue has not already been filed
- [X] I'm reporting the issue to the correct repository (for multi-repository projects)
- [X] I have read and checked all configs (with all optional parts)
- [X] Asked and no solution about my issue with [deepwiki](https://deepwiki.com/kevoreilly/CAPEv2)

# Issue

When using the `reporthtml` feature, there is no behavior for processing in the associated tab. When clicking on the process, the collaspable section just doesn't exist in the HTML, while the `report.json` includes behaviors. In my config I have:

```
[reporthtml]
# Standalone report, not requires CAPE webgui
enabled = yes
# Include screenshots in report
screenshots = yes
apicalls = yes
```

Note that I see screenshots, and previous CAPE version included behaviors inside the `report.html`.

I don't see any failure in logs.

ppil:
Was troubleshooting another issue in the html report, and it seems[ this commit ](https://github.com/kevoreilly/CAPEv2/commit/252b64e38d521766358f5eec2ea5e21c0454ce97)borked it. The http parser now tries to access a non-existent key (req vs requests), then tries to use a non-existent jinja filter (network_rn is not added to filters). After fixing those issues, behavior's empty... looking at that commit, it basically deleted all the behavior reporting code, so it's just not there anymore... :(

kevoreilly:
Fixed in #3157

PR #3139, opened 2026-08-06
Fix broken HTML report/reimplement call table

ncatlin:
### Problem
-----
This PR addresses #3088 where there was errors in HTML report generation and the behaviour section was empty. 
Credit to @zOLakh for reporting it and @ppil for diagnosing it

### Changes
-----
- Fixed the bad req references 
- Added the missing jinja filter
- Added API call table listing to roughly mirror the web version in behaviour and appearence 
    - Includes api category filtering, TID/caller filtering and pagination
    - Does not include the 'search' function because in my testing the web one didn't work either

The HTML reports do get quite big for large traces because unlike the web version we don't have the luxury of calling the chunk API and everything has to be embedded.

### Output
-----

* No more report generation errors 

* Restored the process tree (this isn't actually the same as the more minimal web version)

image

* The call table 
image

### DIFF
--- a/data/html/css/style.css
+++ b/data/html/css/style.css
@@ -43,7 +43,40 @@ body {
 }
 
 .page-link {
-    background-color: #444
+    background-color: #444;
+}
+
+.page-link:hover {
+    background-color: #EE1B2F;
+}
+
+.page-link-active {
+    background-color: #EE1B2F;
+}
+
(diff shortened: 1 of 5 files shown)


ncatlin:
closing as this was on the wrong branch and needs fiddling with

Issue #3138, opened 2026-08-05
The Linux guest install documentation breaks the guest


ncatlin:
Please answer the following questions for yourself before submitting an issue.

- [ N/A] I am running the latest version
- [X] I did read the README!
- [X] I checked the documentation and found no answer
- [X] I checked to make sure that this issue has not already been filed
- [X] I'm reporting the issue to the correct repository (for multi-repository projects)
- [N/A] I have read and checked all configs (with all optional parts)
- [X] Asked and no solution about my issue with [deepwiki](https://deepwiki.com/kevoreilly/CAPEv2)

# Expected Behavior

x64 Linux guests can run UI apps and behave relatively normally

# Current Behavior

The guest installation instructions break the guest UI

## Steps to Reproduce

Following the [documentation](https://capev2.readthedocs.io/en/latest/installation/guest/linux.html#x64-guests) exactly 

* Install a linux guest fresh from a newly downloaded Ubuntu 26 LTS desktop ISO (the docs don't specify a distribution version):
* Take a snapshot - the next commands are going to break your VM
Follow the documentation:
```
$ sudo dpkg --add-architecture i386
$ sudo apt update
$ sudo apt install python3:i386 -y
$ sudo apt install python3-distutils -y  #"'python3-distutils' has no installation candidate"

And now if you 
$ reboot
```
You are dumped at the terminal and can't start the desktop ui. There are no obvious error messages in the syslog or gdm journalctl as to why.

## Context

Note that I am **fully aware that Linux guests are experimental and not officially unsupported** so this issue is not actually a request for support to fix this - but a quick check to see if this is known/expected by anyone who does linux analysis. There do seem to be people using Linux guests judging by recent issues/commits.

It's also an expression of interest for a re-work of Python support:  the dependency on 32-bit python is not great; you can get away with it in the Windows setup because that involves having multiple Python installations at once - but this documented install process fully replaces Python and as time goes on that increasingly makes things break. [Ubuntu discontinued support](https://lists.ubuntu.com/archives/ubuntu-devel-announce/2019-June/001261.html) for 32 bit Python 7 years ago.

Analysis of linux malware is quite a common analysis requirement and IMO is increasing. I don't have a CAPE account so can't see/test linux behaviour on your public instance - but can see from the code that Linux guests don't have a Capemon equivalent, so probably can't offer the rich introspection that Windows guests offer - particularly memory dumps of processes that end quickly. 

[...]

PR #3161, opened 2026-08-15
Update linux_agent.sh

6eh01der:
Some relations to #3138 :

- `apt install python3-pyinotify` instead of `pip3 install pyinotify`
- `tracee` and it's requirements
- fix of incorrect and unneeded `sed`


doomedraven:
not everyone needs tracee so splitted that to different files, thanks

PR #3166, opened 2026-08-18
Ability to disable ratelimit when token authentication is enabled

RobertBendun:
By default it's still turned on, this PR adds ability to turn it of in the custom config via already defined `ratelimit` config entry (that I didn't see used right now and I believe it's intention was to provide this functionality).


doomedraven:
you set user limit in webgui per user

RobertBendun:
In our usage it's better to disable throttling entirely then set high limits per each user or high limits in configs - that's why we propose this change

doomedraven:
make sense, thank you

PR #3171, opened 2026-08-19
Disable stale Remus YARA hash metadata

doomedraven:
Comment out the hard-coded SHA-256 hash in the Remus YARA rules for both the Windows analyzer and CAPE rule set. This avoids stale or misleading metadata being kept in the rule definitions while preserving the rule content and configuration.

They doesn't match those yaras: https://github.com/kevoreilly/CAPEv2/runs/96048388233


kevoreilly:
Damn i forgot it should be 'packed' - thanks

PR #3183, opened 2026-08-21
fix(analyzer): Support 64-bit alignment in defines and add socket tim…

doomedraven:
…eouts to pipe forwarder


doomedraven:
@kevoreilly review when you can https://learn.microsoft.com/es-es/windows/win32/api/tlhelp32/ns-tlhelp32-processentry32. it says -> `ULONG_PTR th32DefaultHeapID;`

Using `ULONG_PTR` ensures that the Python `ctypes` structure **dynamically scales** its size to match the target OS architecture:
* On a 32-bit Python/Windows VM, it will be 4 bytes.
* On a 64-bit Python/Windows VM, it will be 8 bytes.

PR #3206, opened 2026-09-02
Add a generic `attributes` JSON column to machines for flexible per-machine extra data

MMikePL:
Adds a generic `attributes` column to the machines table to store extra per-machine data — whether generic (e.g. last shutdown date) or machinery-specific (e.g. VNC connection details).

The goal is to avoid changing the DB structure for every new field. Any data that doesn't warrant a dedicated column goes into this JSON bag as a key/value pair, so upstream, custom and extension implementations can all add fields  without changing DB structure. This keeps the schema stable across releases and avoids the compatibility problems that frequent structural changes cause for downstream forks.

For optimisation the column is stored as real `jsonb` on Postgres, with a fallback to plain JSON on other back-ends (SQLite/MySQL). `Machine.attributes` is a plain Python `dict` backed by a JSON/JSONB column. Changes are auto-tracked via SQLAlchemy's MutableDict — direct in-place mutation (`machine.attributes["k"] = v`) is fully supported. 
The helpers are still provided as convenience wrappers: `get_attribute(key)` for safe `None`-guard reads, and `set_attribute(key, value)` which also lazy-inits the dict and removes the key when `value=None`.

The column name `attributes` is a proposal — it's a generic  key/value bag, so `properties`, `extra`, `params`, or similar name would work equally well. Open to maintainer preference :)


doomedraven:
this is interesting, i need the same for the tasks so i will review this soon and probaby extend, thank you for the idea!

PR #3219, opened 2026-09-08
tshark

doomedraven:


doomedraven:
Review follow-up pushed to this branch (commit d387eaf87). Verified against `tshark 4.6.6` with pcaps generated byte by byte.

### Blockers fixed

1. **`http.file_data_raw` / `smtp.data.fragment_raw` are not valid field names.** The `_raw` variants only exist in the full-tree output, not as `-e` fields, so tshark exited with status 1 before emitting a single packet and `Pcap2` always returned `{'http_ex': [], 'https_ex': [], 'smtp_ex': []}`. `stderr=DEVNULL` discarded the reason; tshark's stderr is now logged on a non-zero exit.
2. **Bodies were stored as the ASCII hex string.** `http.file_data` is `FT_BYTES`, serialised as hex, and `decode_bytes_field` only unhexlified when a `:` was present. A 23 byte `MZ` payload landed on disk as 46 bytes of hex text, with hashes and preview to match. `FT_BYTES` fields are now unhexlified unconditionally.
3. **Request and status lines were missing.** `http.request.line` / `http.response.line` only carry the header lines. Both are now rebuilt from `http.request.method|uri|version` and `http.response.version|code|phrase`, so the stored header block matches what httpreplay produced and what `_http.html` renders.
4. **Unanswered requests were dropped.** Entries were only emitted from the response branch, so beaconing with no reply, truncated captures and resets disappeared. On the 3-request test pcap only 1 flow survived; pending requests are now flushed at the end of each stream.

### SMTP

`smtp.data.fragment` is `FT_FRAMENUM`, so the old code concatenated frame numbers into the mail body. The DATA payload is now rebuilt from `tcp.payload` between the `DATA` command and its terminating reply (354 does not end the phase), with dot-unstuffing. Also:

- credentials from `smtp.auth.username` / `smtp.auth.password`, base64 decoded, instead of hardcoded `""`
- commands paired with parameters via `smtp.command_line` instead of `zip()`ing two lists of different lengths
- flows reported client to server (the first frame of a stream is the server banner)
- the passlist is applied, which it was not before
- TLS wrapped sessions fall back to the IMF envelope, since `tcp.payload` is ciphertext there

### Other

- `-T ek` streamed line by line instead of buffering the whole JSON document, plus a `tshark_timeout` option (the PR removed the old pcap size guard with no replacement).
- The extra SMTP body pass only runs when a `DATA` command was seen.
- IPv6 flows are no longer dropped (`ipv6.src` / `ipv6.dst`).
- `extract_files = no` no longer advertises hashes for files that were never written (the UI keys its download button off the hash).
- `extract_files` is resolved explicitly: `Dictionary.__getattr__` returns `None` for a missing key, so `getattr(..., True)` never applied its default.
- Duplicate mid-file `import shutil` removed, missing-tshark message downgraded to a warning.

### Tests

The mock-only test is replaced with pcaps built byte by byte and parsed by real tshark, skipped when tshark is absent. 8 passing. Mutation checked: reverting the hex decoding, the request line reconstruction or the unanswered-request flush each fails the suite. The old test also leaked `enabled_passlist = False` into the rest of the session, which is gone.

PR #3221, opened 2026-09-08
fix(web): Close span in process list in network overview

RobertBendun:
Without this fix Network overview tab looks like this:
image

and with this fix:

image


doomedraven:
thank you

PR #3223, opened 2026-09-09
Fix Duplicate indices collision in Wow64 BsonParser/ProtobufParser

doomedraven:


doomedraven:
fixed on capemon side

Issue #3246, opened 2026-09-16
Guacamole Interactive Session Broken: `TypeError` in `uuid3` and Subsequent WebSocket Token Rejection (`token not found in DB`)


ClaudioWayne:
## About accounts on [capesandbox.com](https://capesandbox.com/)
* Issues isn't the way to ask for account activation. Ping capesandbox in [Twitter](https://twitter.com/capesandbox) with your username

## This is open source and you are getting __free__ support so be friendly!

# Prerequisites

Please answer the following questions for yourself before submitting an issue.

- [x] I am running the latest version
- [x] I did read the README!
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I'm reporting the issue to the correct repository (for multi-repository projects)
- [x] I have read and checked all configs (with all optional parts)
- [x] Asked and no solution about my issue with [deepwiki](https://deepwiki.com/kevoreilly/CAPEv2)

# Expected Behavior

When clicking on the remote session / interactive desktop feature for a task the interactive Guacamole session should load successfully in the browser and establish a WebSocket connection.

# Current Behavior

1. **Initial Exception (`TypeError`):**
When clicking on the remote session / interactive session link it fails with a `TypeError: can't concat int to bytes` in `uuid3`:
```text
File "/opt/CAPEv2/web/submission/views.py", line 959, in remote_session
  session_id = uuid3(NAMESPACE_DNS, task_id).hex[:16]
File "/usr/lib/python3.12/uuid.py", line 718, in uuid3
  namespace.bytes + name,
TypeError: can't concat int to bytes

```
2. **WebSocket Error after Workaround:**
Modifying line 957 and 920 in `web/submission/views.py` to convert `task_id` to a string (`session_id = uuid3(NAMESPACE_DNS, str(task_id)).hex[:16]`) resolves the initial exception. However, the UI then fails with a connection error overlay:
`Connection error: An unexpected error occurred: undefined`
Checking `guac-server.log` (created by web/guac/consumers.py) reveals that the WebSocket connection is rejected:
```text
WARNING:guac-session:WebSocket rejected: token not found in DB

```
In the browser developer tools, the WebSocket connection fails with `NS_ERROR_WEBSOCKET_CONNECTION_REFUSED`.

# Failure Information (for bugs)

* **Attempted fixes & PRs:**
* Applied the patch from [PR #3217](https://github.com/kevoreilly/CAPEv2/pull/3217).
[...]

PR #3249, opened 2026-09-21
fix Guacamole Interactive Session

ClaudioWayne:
fix TypeError in uuid3 and Subsequent WebSocket Token Rejection (token not found in DB)
-> https://github.com/kevoreilly/CAPEv2/issues/3246


doomedraven:
Thanks

